PR #2, opened 2016-04-07
Added configuration class

woodpigeon:
Add configuration to allow a consuming app to configure the engine e.g. 
```
FloodRiskEngine.configure do |config|
  config.address_facade_server = "some value"
end
```

### DIFF
--- a/lib/flood_risk_engine/configuration.rb
+++ b/lib/flood_risk_engine/configuration.rb
@@ -1,0 +1,30 @@
+# Class for setting configuration options in this engine. See e.g.
+# http://stackoverflow.com/questions/24104246/how-to-use-activesupportconfigurable-with-rails-engine
+#
+# To override default config values, for example in an initaliser, use e.g.:
+#
+#   FloodRiskEngine.configure do |config|
+#    config.exemptions_expire_after_duration = 3.years - 1.day
+#   end
+#
+# To access configuration settings use e.g.
+#   FloodRiskEngine.config.exemptions_expire_after_duration
+#
+module FloodRiskEngine
+  class Configuration
+    include ActiveSupport::Configurable
+
+    # Define accessors here, e.g.
+    # config_accessor(:some_key)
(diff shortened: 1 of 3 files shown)


woodpigeon:
@reggieb config class for when we need it, feel free to amend

lib/flood_risk_engine/configuration.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# To access configuration settings use e.g.
+#   FloodRiskEngine.config.exemptions_expire_after_duration
+#
+module FloodRiskEngine
+  class Configuration
+    include ActiveSupport::Configurable
+
+    # Define accessors and optional defaults
+    config_accessor(:test_me) { "test" }
```
reggieb: Should this be a commented out example?

spec/lib/configuration_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module FloodRiskEngine
+  describe Configuration do
+    it "adds #config to the engine module" do
+      expect(FloodRiskEngine).to respond_to(:config)
+    end
+
+    it "has the correct default values" do
+      expect(FloodRiskEngine.config.test_me).to eq "test"
+    end
```
reggieb: I'd remove this test as the 'is configurable via a block' test tests the functionality without relying on the presence of special - only for testing - configuration item.

reggieb: Or change the test to so it test the expected response when you call a non-existent config item.

Actually, that's the better option.

woodpigeon: @reggieb I should have made it clear in the code that test_me was designed to be replaced by the first real config value and was just there to set a precedent - but that was not a poor idea so have made it an example as you suggest. Now test for non-existence as you suggest too - thx

autotelik:
:ship:

reggieb:
:ship: it

PR #3, opened 2016-04-07
Add ea-address_lookup gem

woodpigeon:
Add a reference to the ea-address_lookup gem in preparation for postcode and uprn lookups.


woodpigeon:
Adding the gem will be done as part of #66

PR #7, opened 2016-04-13
Views based on prototype version 1

autotelik:
Create views  that fit the pattern of header and step form  layout

View names to be based on the associated state names  from prototype version 1


autotelik:
:eyes:

PR #13, opened 2016-04-14
Feature/add work flow object and back flows

reggieb:
The new WorkFlow object defines the sequence of steps used by a journey controlled by the state machine.

Adding the WorkFlows has facilitated the additions of a `go_back` event that uses the reverse work flow of the `next_step`  event.

Detaches StepMachine from Enrollment, so the same state machine can be used by any object that needs state logic.

Renames `next_step` event as `go_forward`

**NOTE - the controller `update` method has been commented out to get the commit through rubocop**


woodpigeon:
:shipit:

PR #17, opened 2016-04-15
Feature/first good journey

reggieb:
This pull request sets up the engine so that it will complete a good journey.

Things to note:

- I ended up splitting the WorkFlow out of the state engine.  When I started entering the work flows, having to repeat each value as the key to the next pair became a pain. When I thought about it, the work flow is really just an array anyway. So having a class that let me define the work flows as arrays, and then did the conversion to state machine friendly hashes made sense to me.
- I've kept the existing actions where I could.
- All the "new" actions
  - have the simplest of view partials added as place holders.
  - use the NullForm (which I've updated so that it works without any input)
  - use a generic i18t locale at `config/locales/flood_risk_engine/enrollments/steps/edit/generic.en.yml
- I couldn't get the existing user_type template to work - so I've removed it for now. I'm assuming one from copy kit can be used to put it back when needed.


app/state_machines/flood_risk_engine/work_flow.rb
```diff
@@ -0,0 +1,71 @@
...
+      def local_authority
+        start + local_authority_branch + finish
+      end
+
+      def start
+        [
+          :check_location,
+          :check_exceptions,
+          :add_exceptions,
+          :grid_reference,
```
woodpigeon: add_exemptions and check_exemptions

woodpigeon: reverse add_exemptions and check_exemptions

app/state_machines/flood_risk_engine/work_flow.rb
```diff
@@ -0,0 +1,71 @@
...
+          :user_type
+        ]
```
woodpigeon: I used organisation_type instead if user_type - there is a working organisation_type_form ready but I could be renamed to user_type_form. We should probably have a go at nailing down what we want to call the steps (independent of the prototype) sooner rather than later so the step names don't confuse other developers down the line.

app/state_machines/flood_risk_engine/work_flow.rb
```diff
@@ -0,0 +1,71 @@
...
+    # Converts [:a, :b, :c] into
+    # {:a => :b, :b => :c}
+    def make_hash(array)
+      all_but_last = array[0..-2]
+      all_but_first = array[1..-1]
+      pairs = [all_but_last, all_but_first].transpose
+      Hash[pairs]
+    end
+  end
+end
```
woodpigeon: I like way your Workflow class has panned out.

app/views/flood_risk_engine/enrollments/steps/_add_exceptions.html.erb
```diff
@@ -0,0 +1 @@
+add_exception
```
woodpigeon: filename add_exemptions

app/views/flood_risk_engine/enrollments/steps/_check_exceptions.html.erb
```diff
@@ -0,0 +1 @@
+check_exceptions
```
woodpigeon: filename check_exemptions

autotelik:
just a general comment  the concepts and files seems to be growing  .. states, steps and now workflow. I though we were trying to make things simpler ?  its not particularly clear to me what the difference is between the terms and why we need anything other than  'state'

reggieb:
@autotelik I'll label up the classes so it's a little clearer what each one does.

reggieb:
I found a couple of bugs, which I've now fixed.

autotelik:
For reference the Forms coming out of Copykit now follow the Null pattern so they can be dropped in and wiil work out of the box, and then be finessed as we do the page by page stories. So form_object_class_map could be made dynamic and not hard coded

autotelik:
:ship:

PR #18, opened 2016-04-15
Remove simple_form

woodpigeon:
Remove the simple_form gem and its associated files as it is not going to be needed.


autotelik:
:ship:

reggieb:
:ship: it

PR #22, opened 2016-04-17
Add missing db indexes and relationships

woodpigeon:
- Add database indexes
- Remove dummy enrollment columns from the prototyping phase
- Add `dependent:` to has_*
- Removed `accepts_nested_attributes_for` calls as our form object approach has hopefully removed the need for them


autotelik:
:ship:

reggieb:
:ship: it

PR #23, opened 2016-04-18
Create page : check location - RIP-53

autotelik:
https://eaflood.atlassian.net/browse/RIP-53

This PR does NOT complete the Story - Lots of change and setup in here so keeping PR smaller and getting into develop

Add view for the check location page
Add in a locale file for all text elements on the check location page
Remove defunct users DB from database.yml
Fix migration issues with Postgres (calls that work in sqlite but not postgres)
Create form object and view for Check Location page
Create static page with EA Contact details for No selection
Add error handling view helpers to create anchors and error text boxes


app/controllers/flood_risk_engine/application_controller.rb
```diff
@@ -2,6 +2,8 @@ module FloodRiskEngine
   class ApplicationController < ActionController::Base
     protect_from_forgery with: :exception
 
+    layout ->(_) { 'application' }
+
```
woodpigeon: Just wondering - should the engine be specifying the layout the host app will use? Just compared to devise and its agnostic about layout

db/migrate/20160413121293_create_exemptions.rb
```diff
@@ -21,5 +21,10 @@ def change
 
     add_foreign_key :flood_risk_engine_enrollments_exemptions,
                     :flood_risk_engine_exemptions, column: :exemption_id
+
+    add_index :flood_risk_engine_enrollments_exemptions,
+              [:enrollment_id, :exemption_id],
+              unique: true,
+              name: "fre_enrollments_exemptions_enrollment_id_exemption_id"
   end
```
reggieb: Once a migration has been committed to the repository it shouldn't be modified, unless something is broken.

app/controllers/flood_risk_engine/enrollments_controller.rb
```diff
@@ -2,11 +2,49 @@
...
+      end
+
+    end
+
+    private
+
+    # The form object factory dont work for new Enrollment sso have to roll our own
+    def form_object
+      "FloodRiskEngine::Steps::#{@enrollment.current_step.classify}Form".constantize.factory(@enrollment)
+    end
```
woodpigeon: As this isn't really a step might it be worth moving the form object from `forms/flood_risk_engine/steps` to `forms/flood_risk_engine` and then just accessing it as FloodRiskEngine::CheckLocationForm?

app/controllers/flood_risk_engine/application_controller.rb
autotelik: Yeah can do. AFAIK this engine will only ever be used for this App so going to great lengths to make it act like a generic engine not my priority

config/routes.rb
```diff
@@ -1,6 +1,10 @@
 FloodRiskEngine::Engine.routes.draw do
+  resources :enrollments
+
+  # Copied from dummy app - may be out of date now
   resources :enrollments, only: [:new] do
```
reggieb: The enrollment resources should only be defined once. Rather than adding an additional `resources :enrollments` the existing one should be modified to added the required additional actions. So:

```
  resources :steps, only: [:show, :update], controller: "enrollments/steps"
end
```

db/migrate/20160413121293_create_exemptions.rb
autotelik: it was broken in postgres

app/models/flood_risk_engine/enrollment.rb
```diff
@@ -42,6 +42,10 @@ def state_machine
       to: :state_machine
     )
 
+    def state
+      self.step
+    end
+
```
woodpigeon: alternative `alias_attribute :state, :step` - but if we start mixing up state and step (we will also have a status eg active, deregistered etc) then the code could be become harder to follow

app/views/flood_risk_engine/enrollments/steps/_check_location.html.erb
```diff
@@ -1 +1,31 @@
...
+      </label>
+
+      <label class="block-label" for="location_check_no">
+        <%= f.radio_button :location_check, :no, { checked: false, :class => "radio", id: "no_option" } %>
+        <%= t '.label1' %>
+      </label>
+  <% end %>
+  </div>
+
+</fieldset>
```
woodpigeon: as this is not a step could move out of steps folder

app/controllers/flood_risk_engine/enrollments_controller.rb
autotelik: feels like a step to me

config/routes.rb
autotelik: didn't know what the nested stuff was supposed to be about so just did that to get going .. seemed to work fine but changed to above

config/routes.rb
```diff
@@ -1,6 +1,10 @@
 FloodRiskEngine::Engine.routes.draw do
+  resources :enrollments
+
```
woodpigeon: `resources :enrollments` already defined below.
maybe change the one below to `resources :enrollments, only: [:new, :create] do`

app/views/flood_risk_engine/enrollments/steps/_check_location.html.erb
autotelik: it is a step

  class EnrollmentStateMachine < FiniteMachine::Definition
    initial :check_location

app/models/flood_risk_engine/enrollment.rb
autotelik: right

app/views/flood_risk_engine/enrollments/steps/_check_location.html.erb
reggieb: The steps need to be updated so that this isn't the initial step.

woodpigeon: OK, I just thought this initial page (cos it featured a redirect) was going to be enrollments#new and outside of the state machine. If its it is a step then maybe we need to find a way for a form object to return a redirect url on #save, then it can be a proper step. I'll have a play.

reggieb: It's only a step because it's defined in the state machine workflow. All it takes is to remove `:check_location` from `WorkFlow` and set the initial step in `EnrollmentStateMachine` to `:add_exemptions`

autotelik: Continue button goes to create, sets the step as having been completed and redirects to the steps controller. It all works quite nicely but go for it.

autotelik: Ok my thinking was, Once it gets to the StepController the initial step has been saved in the DB as check_location, indicating that Yes the Location has been checked and the journey under way.   I thought the Step saved in the DB was a step that has already been completed, this way it seems more clear the jounrey has started rather than creating an Enrollment with a nil state

It all works fine in the front end

app/models/flood_risk_engine/enrollment.rb
reggieb: Actually, the easiest way is to add it to the list of methods passed through via delegate.

autotelik:
Please either merge this or explain clearly how you want me to code this as I am not sure what the objections are really about, or how you want this page integrated into the new grand scheme of things. Currently choosing Yes acts to create an Enrollment, i.e the journey has started, and then the user is redirected to the StepController to the url and the add_exemptions stub page is shown

The start page leads to the Check location page, Validation works, No directs to a static page,  Yes to add_exemptions.

app/views/flood_risk_engine/enrollments/steps/_check_location.html.erb
reggieb: Step is persisted on save and therefore also on create:

```
FloodRiskEngine::Enrollment.create
   (0.1ms)  begin transaction
  SQL (0.5ms)  INSERT INTO "flood_risk_engine_enrollments" ("step", "created_at", "updated_at") VALUES (?, ?, ?)  [["step", "check_location"], ["created_at", "2016-04-21 08:55:23.293027"], ["updated_at", "2016-04-21 08:55:23.293027"]]
   (135.1ms)  commit transaction
 => #<FloodRiskEngine::Enrollment id: 1, applicant_contact_id: nil, created_at: "2016-04-21 08:55:23", updated_at: "2016-04-21 08:55:23", step: "check_location", organisation_id: nil, site_address_id: nil> 
```

autotelik:
One last reason to leave it as it as - saving it a a step, the back link works ..i.e takes you back to render http://localhost:3000/fre/enrollments/48/steps/check_location

reggieb:
:ship: it

PR #24, opened 2016-04-18
Feature/add step control to steps controller

reggieb:
Ensures that the step in the url matches `enrollment.step`. It also allows users to move back through the journey.

Also refactors routes so that `enrollment_step_path(enrollment, step)` can be used to get to each step. To facilitate this, the main step action has changed from `edit` to `show`


autotelik:
If simplification was the goal we seem to have far more code, classes and methods than DSC . That's my only comment  so :ship: from me

PR #25, opened 2016-04-20
Allow a user to define organisation type at user_type step

reggieb:
Use an ActiveRecord `enum` to define organisation types, as validation has moved to the form object.

### DIFF
--- a/app/forms/flood_risk_engine/steps/user_type_form.rb
+++ b/app/forms/flood_risk_engine/steps/user_type_form.rb
@@ -1,8 +1,8 @@
 module FloodRiskEngine
   module Steps
     class UserTypeForm < BaseForm
-      property :type
-      validates :type, presence: true
+      property :org_type
+      validates :org_type, presence: true
 
       def self.factory(enrollment)
         organisation = enrollment.organisation || Organisation.new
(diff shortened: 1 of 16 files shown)


autotelik:
is this sufficient to capture the major differences between organisation types ?

for example

class Partnership < DigitalServicesCore::Organisation
      has_many :partner_contacts

Are whatever  savings you are making in the state machine worth it, considering we are once again invalidating everything from WEX  models, page objects, factories, tests  ?

reggieb:
The partnership is the only instance I can think of when there is a difference between organisation types. As this app doesn't validate at the model level, all the benefits of using a enum field outweighs the issue of having to deal with this one case IMO.

autotelik:
what benefits out weigh the flexibility of proper OO modelling ? 

This was the code to create all the next evens for the OrgType branch .. can't see how an enum instead of classes would make that shorter or more understandable
```
    WasteExemptionsShared::OrganisationTypes.types_display_list.each_with_index do |t, i|
      org_type_state = WasteExemptionsShared::OrganisationTypes.type_names_list[i].to_sym
      transition from: :choosing_organisation_type, to: org_type_state, on: :next,
                 if: ->(e) do
                   e.organisation && e.organisation.sti_class == t
                 end
    end
```

but since your doing the branching code, go for it :ship:

woodpigeon:
I'm happy to go this way for now as it removes the STI complexity and we'll have use a simpler approach to adding the partnership relationship which is the only thing this approach leaves unsolved :ship:

reggieb:
rebased

PR #30, opened 2016-04-22
Add a FloodRiskEngine initialiser

woodpigeon:
This will be the central location for configuring the engine.


woodpigeon:
Added in error

PR #31, opened 2016-04-25
Create user_type/local_authority - RIP-73

autotelik:
Add the Form
Add the partial view


autotelik:
:eyes:

app/validators/flood_risk_engine/name_format_validator.rb
```diff
@@ -0,0 +1,18 @@
...
+  class NameFormatValidator < ActiveModel::EachValidator
+
+    def validate_each(record, attribute, value)
+      if(value !~ NameFormatValidator.valid_name_regex)
+        record.errors.add attribute,  I18n.t("flood_risk_engine.validation_errors.name.invalid")
+      end
+    end
+
+    def self.valid_name_regex
+      /\A[\p{Alpha} ,\.'-]+\z/i
```
reggieb: I think `/\A[A-Za-z ,\.'-]+\z/i` is a little clearer.

autotelik: I prefer to stay consistent with WEX and afaik Alpha was originally chosen as your version would fail for e.g Björk Guðmundsdóttir

app/factories/flood_risk_engine/steps/form_object_factory.rb
```diff
@@ -38,11 +38,16 @@ def form_object_class_map
...
 
         def setup_form_object(step)
-          "FloodRiskEngine::Steps::#{step.to_s.classify}Form".constantize
-        rescue
-          Rails.logger.debug(" No Form Object found for step [#{step}] - falling back to object_class map")
-          nil
+          form_name = "FloodRiskEngine::Steps::#{step.to_s.classify}Form"
+          begin
+            form_name.constantize
+          rescue => x
```
reggieb: If we are going down the meta route for determining the form name, wouldn't it be easier to check `form_object_class_map` first and then try to work out the form class from the step name if the step isn't in the list? Then we wouldn't need this rescue clause.

Then as we add named forms we can remove the respective step from `form_object_class_map`. Ultimately, `form_object_class_map` will empty and at that point we can drop it.

If the exception catching is to stay then it should be limited to only catch the error generated because the constant doesn't exist. That is:

```
  rescue NameError => x
```

app/factories/flood_risk_engine/steps/form_object_factory.rb
```diff
@@ -38,11 +38,16 @@ def form_object_class_map
         private
 
         def setup_form_object(step)
-          "FloodRiskEngine::Steps::#{step.to_s.classify}Form".constantize
```
reggieb: I think `"#{step.to_s.classify}Form".constantize` would work as we already in the `FloodRiskEngine::Steps` name space

app/factories/flood_risk_engine/steps/form_object_factory.rb
```diff
@@ -38,11 +38,16 @@ def form_object_class_map
...
-        rescue
-          Rails.logger.debug(" No Form Object found for step [#{step}] - falling back to object_class map")
-          nil
+          form_name = "FloodRiskEngine::Steps::#{step.to_s.classify}Form"
```
reggieb: I think `form_name = "#{step.to_s.classify}Form".constantize` would work as we already in the `FloodRiskEngine::Steps` name space

app/validators/flood_risk_engine/name_format_validator.rb
reggieb: OK.

reggieb:
If you change line 44 of form_object_factory.rb to:

```
rescue NameError => x
```

I'm happy for this to be :ship:ped

app/forms/flood_risk_engine/steps/local_authority_form.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      validation :name_present? do
+        validates :name, presence: {
+          message: I18n.t("#{LocalAuthorityForm.locale_key}.errors.name.blank")
+        }
+      end
+
+      include ActiveModel::Validations
+      validation :email_format, if: :name_present? do
```
woodpigeon: There is an initializer `flood-risk-engine/config/initializers/reform.rb` which includes `Reform::Form::ActiveModel::Validations` - but maybe not sufficient?

app/forms/flood_risk_engine/steps/local_authority_form.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      def self.name_max_length
+        200
+      end
+
+      validation :name_present? do
+        validates :name, presence: {
+          message: I18n.t("#{LocalAuthorityForm.locale_key}.errors.name.blank")
+        }
+      end
```
woodpigeon: is validation block needed? Does the simpler alternative work?
``` validates(
        :name,
        presence: {
          message: I18n.t("#{LocalAuthorityForm.locale_key}.errors.name.blank")
        }
      )
```

app/factories/flood_risk_engine/steps/form_object_factory.rb
autotelik: as it's early days  I was just trying to do the dynamic thing without treading on your toes if you wanted to use case  - so I'm happy to tweak this  as everyone sees fit

app/forms/flood_risk_engine/steps/local_authority_form.rb
woodpigeon: just wondering if this is works 
```
validates :name, 'flood_risk_engine/name_format': true, allow_nil: true
```

autotelik: yes pulls in custom validators ... I wasn't sure whether to put that in base form ?

I'm not sure what allow_nil: true  means ?  the scheme as it stands uses reform validation groups to ensure the format validator is not run iif the name is blank, so the errors reported to user are quite nice and specific

woodpigeon: is `email_format` a copy and paste thing :smile:

app/views/flood_risk_engine/enrollments/steps/_local_authority.html.erb
```diff
@@ -1 +1,14 @@
-local_authority
+<fieldset>
+
+  <legend class="visuallyhidden"><%= t('.legend') %></legend>
+
+  <%= form_group_and_validation(form, :name) do %>
+      <div class="form-group">
+        <label class="form-label" for="authority-name"><%= t('.label') %></label>
+        <p class="form-hint"><%= t('.para_text') %></p>
```
woodpigeon: Nice. Shouldn't we use the more Railsy '<%= f.label_for .. %>'  to guarantee labels are always associated with the right input eg `<%= f.label :name, t('.label'), class: 'form-label' %>`?

config/locales/flood_risk_engine/_local_authority.en.yml
```diff
@@ -0,0 +1,14 @@
...
+      steps:
+        local_authority: 
+          heading: What's the name of the local authority or public body?
+          label: Local authority or public body name
+          legend: Local authority or public body name
+          para_text: For example, Bristol City Council
+          errors:
+            name:
+             blank: Enter the name of the local authority or public body
+             too_long: Check the name - it should have no more than than 200 characters
```
woodpigeon: I like having the errors in the step yml like this - everything in one place, easy to find

app/forms/flood_risk_engine/steps/local_authority_form.rb
woodpigeon: [allow_nil](http://guides.rubyonrails.org/active_record_validations.html#allow-nil) (or maybe [allow_blank](http://guides.rubyonrails.org/active_record_validations.html#allow-blank) better?) I didn't know about reform [validation groups](http://trailblazer.to/gems/reform/validation.html#validation-group) but the cynic in may says other devs are likely to think its a rails thing and scratch their heads a bit - I'd rather lean on pure Rails validations if poss without introducing too much reform-specifc magic :tophat:

config/locales/flood_risk_engine/_local_authority.en.yml
reggieb: The down side is that ActiveModel doesn't know where to find the error texts, so you have to specify the whole path to the message. If you have a section:
```
  activemodel:
    errors:
      messages:
        some_error: Some error message
```
You can then do:
```
validates(
        :name,
        presence: {
          message: I18n.t("some_error")
        }
      )
```
I guess that could live in another part of the other locale data for the page/form.

app/factories/flood_risk_engine/steps/form_object_factory.rb
autotelik: this does not seem to be the case, constantize starts to fail if I switch to "#{step.to_s.classify}Form"

app/factories/flood_risk_engine/steps/form_object_factory.rb
autotelik: changes made as per comment - order changed, rescue tightened up

reggieb: I think the simplest solution is just:
```
klass = form_object_class_map[step.to_sym] || setup_form_object(step)
klass.factory(enrollment)
```
Then setup_form_object can become:
```
def setup_form_object(step)
  "#{step.to_s.classify}Form".constantize
end
```

app/forms/flood_risk_engine/steps/local_authority_form.rb
autotelik: ok .. apart from the fact that

   validation :validate_name_format, if: :name_present? do

seems  much more understandable to me than  allow_nil ? 

Having validatins that can be chained like this seems an important step to solving many of the wex pain

app/views/flood_risk_engine/enrollments/steps/_local_authority.html.erb
autotelik: Agreed.

app/forms/flood_risk_engine/steps/local_authority_form.rb
autotelik: as above ... this allows the validation result to be used elsewhere allowing the more readable

```validation :validate_name_format, if: :name_present? do```

app/factories/flood_risk_engine/steps/form_object_factory.rb
reggieb: Strange. OK - looks like I'm wrong then.

autotelik: I really did try it - honest guv :-)

autotelik:
:ship: I think everything  addressed that needs addressing - please check and either ship or let me know issues .. tvm

autotelik:
:eyes:

autotelik:
@woodpigeon  <boing due to sprint retro>

PR #32, opened 2016-04-25
RIP-60 add add_exemptions functionality

reggieb:
Allows multiple exemptions to be selected per journey


autotelik:
does not look like the views will match the content doc and there is no locale information containing the text elements of the pages.

autotelik:
when the validation fails how is the expected error messages displayed to the user ?  expect to see some kind of anchors from the error box to the field with the error

autotelik:
p.s there is a helper you can use to just wrap the element which gives you the linkage back to the main error box :  see  <%= form_group_and_validation(form, :base) do %>

app/forms/flood_risk_engine/steps/add_exemptions_form.rb
```diff
@@ -1,27 +1,29 @@
...
+
       def self.factory(enrollment)
         new(enrollment)
       end
 
+      validates(
+        :exemption_ids,
+        length: {
+          minimum: 1,
+          message: "Select at least one exemption"
```
autotelik: error messages should also be stored in locales

spec/forms/flood_risk_engine/add_exemptions_form_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        expect(enrollment.exemptions).to eq(exemptions)
+      end
+    end
+
+    describe ".validate" do
+      context "with empty exemptions params" do
+        let(:params) { { params_key => { exemption_ids: [] } } }
+        it "should fail" do
+          expect(subject.validate(params)).to be(false)
+        end
```
woodpigeon: Could maybe assert that the correct validation message is there (as specced in prototype powerpoint) ie loaded from a translation file

spec/forms/flood_risk_engine/add_exemptions_form_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      context "with no exemptions params" do
+        let(:params) { { params_key => {} } }
+        it "should fail" do
+          expect(subject.validate(params)).to be(false)
+        end
+      end
+
```
woodpigeon: as above

woodpigeon:
@reggieb @autotelik and I have added inline comments.

reggieb:
The form is a single set of check boxes and there is only one error - you haven't selected one of the check boxes. So I'm not really sure how you anchor the error to a particular element on this page. I've wrapped the checkboxes in a `form_group_and_validation` as I expect the formatting will go astray if I don't.

reggieb:
The page content is gathered from the exemption's properties, so other than the error message, there is nothing to add to the locales.

app/forms/flood_risk_engine/steps/add_exemptions_form.rb
reggieb: Good point. I've moved it to the locales.

autotelik:
ok I might have been looking at wrong prototype version as was expecting to see some header text and some footer text

spec/forms/flood_risk_engine/add_exemptions_form_spec.rb
reggieb: I've added a check at line 38

reggieb:
I think I haven't got the page title displaying correctly - I should fix that shortly.

autotelik:
its not clear to me how the header text "You can register more than one exemption but they must all be at the same location. You should check you can meet the conditions for each exemption. Exempt flood risk activities guidance."  ... is inserted  into the layout and stored in locale ?

reggieb:
Sorry @autotelik  - my mistake - I see what you mean.

reggieb:
Rebased

PR #33, opened 2016-04-27
RIP-62 check your exemptions

reggieb:
https://eaflood.atlassian.net/browse/RIP-62

Configure page where users can review the exemptions they have added.

### DIFF
--- a/app/controllers/flood_risk_engine/enrollments/exemptions_controller.rb
+++ b/app/controllers/flood_risk_engine/enrollments/exemptions_controller.rb
@@ -1,0 +1,32 @@
+require_dependency "flood_risk_engine/application_controller"
+
+module FloodRiskEngine
+  class Enrollments::ExemptionsController < ApplicationController
+
+    def show
+      destroy # allows exemptions to be removed when JavaScript is disabled.
+    end
+
+    def destroy
+      enrollment.exemptions.destroy(exemption)
+      step_back if enrollment.exemptions.empty?
+      redirect_to enrollment_step_path(enrollment, enrollment.current_step)
+    end
+
+    private
+
+    def enrollment
+      @enrollment ||= Enrollment.find(params[:enrollment_id])
+    end
+
+    def exemption
+      @exemption ||= Exemption.find(params[:id])
+    end
+
+    def step_back
+      enrollment.go_back
+      enrollment.save
+    end
+
+  end
+end
--- a/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
+++ b/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
@@ -11,7 +11,7 @@ module FloodRiskEngine
     class StepsController < ApplicationController
       class StepError < StandardError; end
       rescue_from StepError, with: :step_not_found
-      before_action :check_step_is_valid
+      before_action :check_step_is_valid, except: [:remove_exemption]
       before_action :back_button_cache_buster
 
       def show
@@ -30,10 +30,18 @@ module FloodRiskEngine
         end
       end
 
+      def remove_exemption
+        @enrollment = Enrollment.find(params[:id])
+        @exemption = Exemption.find(params[:exemption_id])
+        enrollment.exemptions.destroy(@exemption)
+        step_back if enrollment.exemptions.empty?
+        redirect_to step_url
+      end
+
       private
 
       def step
-        params.fetch(:id).to_sym
+        @step ||= params.fetch(:id).to_sym
       end
 
       def step_url
--- a/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
+++ b/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
@@ -30,14 +30,6 @@ module FloodRiskEngine
         end
       end
 
-      def remove_exemption
-        @enrollment = Enrollment.find(params[:id])
-        @exemption = Exemption.find(params[:exemption_id])
-        enrollment.exemptions.destroy(@exemption)
-        step_back if enrollment.exemptions.empty?
-        redirect_to step_url
-      end
-
       private
 
       def step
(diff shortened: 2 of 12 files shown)


app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -30,10 +30,18 @@ def update
         end
       end
 
+      def remove_exemption
+        @enrollment = Enrollment.find(params[:id])
```
autotelik: @woodpigeon  In WEX we had to implement token based lookup .. is it worth getting that in now  to save factoring calls like  find(params[:id])   later ?

autotelik: Even if we add token stuff  later is it worth replacing raw calls to `Enrollment.find(params[:id])` with a `set_enrollment` type method to make it easier to refactor later if needed ?

config/routes.rb
```diff
@@ -1,5 +1,14 @@
 FloodRiskEngine::Engine.routes.draw do
   resources :enrollments, only: [:new, :create] do
     resources :steps, only: [:show, :update], controller: "enrollments/steps"
+
+    member do
```
autotelik: Seems to be polluting the state engine with functionality that should belong in a different controller

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -30,10 +30,18 @@ def update
...
+        @exemption = Exemption.find(params[:exemption_id])
+        enrollment.exemptions.destroy(@exemption)
+        step_back if enrollment.exemptions.empty?
+        redirect_to step_url
+      end
+
```
woodpigeon: Two thoughts on alternatives to this approach: 

**1 All step-based knowledge mored to a new  CheckExemptionsForm form object**

- The 'Remove' links use a `link_to(method: :patch)` (clicking 'Remove' changes something on the server so arguably should be PATCH (or, actually, DELETE) rather than a GET). However might need to check non-js compatibility (`method: :x` on `link_to` is an unobtrusive js thing so might have to hand crank an html form around each link) 
- add a CheckExemptionsForm for this step (which would incidentally maintain the 'form object for every step' developer expectation, reducing cognitive friction. I love that phrase).
Override the form object #save.
`CheckExemptionsForm#save` would have to handle 2 situations
  1. If Remove clicked, an exemption_id is supplied - remove the exemption and redirect to same page, or redirect to add_exemptions if none left
  2. If Continue clicked, no exemption_id is passed so an assumtion made that this is the normal 'next' action - essentially nothing to do in this instance as nothing to save.

Like this, steps controller remains unchanged, and all the logic about a step is in the form object,
even if its a bit clumsy for this step. I am just concerned that otherwise we set a precedent for putting step-specific logic in the controller, offering a misleading path of least resistence for
future devs working on the code


**2 Enrollments::ExemptionsController**

Even better maybe is another alternative: to add an `Enrollments::ExcemptionsController` with a single `#delete` action,  and the Remove links hits this - which is a bit more HTTP compliant 
and RESTful than a PATCH for this case. 
However this will spread the step logic across an extra controller - not sure if that's more confusing or not

autotelik: >>non-js compatibility

rings a bell from wex ... I believe you can use a  `button_to method: :delete`  to replace link_to when javascript turned off  ...  and just style the button so it looks like a link

reggieb: Yes - I started with using a patch method on the link, but then realised that if JavaScript was turned off, it would become a GET. So rather than handling that as a separate case, I made it the main behaviour.

reggieb: I think the Enrollments::ExemptionsController option is better than the form object one.

config/routes.rb
reggieb: State machine isn't involved - this is the steps controller. 

My feeling was that for one action, it wasn't worth the extra work of creating another controller. However, I'm happy to be persuaded otherwise (which seems to be the case from the discussion below)

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
reggieb: I'd move it into a `set_enrollment` method at the first time it was needed in another method. Also I avoided this approach here as there is already the `enrollment` method doing the same job in this controller.

All in all - this is further reason for moving the functionality into another controller.

reggieb:
Thanks you both for the discussion. I agree with you both - this needs to be stripped out into another controller.

reggieb:
Unfortunately the rebase has removed some of the original comments.

app/controllers/flood_risk_engine/enrollments/exemptions_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def destroy
+      enrollment.exemptions.destroy(exemption)
+      step_back if enrollment.exemptions.empty?
+      redirect_to enrollment_step_path(enrollment, enrollment.current_step)
+    end
+
+    private
+
+    def enrollment
```
autotelik: given its a dynamic language is there a slight performance hit using this form of always calling a method and using the cached result ? i.e rather than assigning once to a @enrollment  variable, locally  within the method ?   ..  on surface appears each subsequent call to enrollment requires couple more operations than simply using @enrollment 

Just a general question rather than a comment/suggestion (in big scheme of network etc sure its irrelevant)  .. I'm not sure how the underlying ruby compiler could/would optimise this

autotelik:
:ship:  from me

app/controllers/flood_risk_engine/enrollments/exemptions_controller.rb
reggieb: There is a very small cost to calling a method, but it is very small. I haven't performance tested it, but am confident the cost is smaller than you can accurately measure.

I use this pattern a lot, because it means I don't have to worry about where and when the object is first initialised. It also ensures that the object is only initialised when it is used, and that it is only initialised once. 

It is a lazy load technique, but I'm confident my code has got cleaner since using it.

autotelik: yeah its a good technique  ... and sure the cost is not worth worrying about .. was just curious

PR #34, opened 2016-04-27
Allow db seeding on certain patforms

woodpigeon:
Only run seeds if we are not in production, unless an
ALLOW_PRODUCTION_SEEDING ENV var has a value 'true'. This lets us
enable seeding on certain test deployment platforms, for example
Heroku, where accidental seeding is not an issue.


woodpigeon:
On hold - better solution suggested by @reggieb

woodpigeon:
A better solution was introduced using dibber in another PR

PR #35, opened 2016-04-27
Use Dibber to manager data seeding

reggieb:
Dibber will handle the gathering of data from source YAML files and inserting that data into the database.

If an object doesn't exist, Dibber will insert it into the database. If the object does exist, Dibber will update the object.

See https://github.com/reggieb/Dibber


db/seeds.rb
```diff
@@ -1,13 +1,16 @@
...
-# and the relevant environment specific file. This solution and the code below were taken from
-# https://archive.dennisreimann.de/blog/seeds-for-different-environments/
-['all', Rails.env].each do |seed|
-  seed_file = FloodRiskEngine::Engine.root.join('db', 'seeds', "#{seed}.rb")
-  if File.exist?(seed_file)
-    puts "FRE:SEED MSG: Loading #{seed} data"
-    require seed_file
-  end
-end
+require 'dibber'
```
autotelik: is this Env specific ?

db/seeds.rb
```diff
@@ -1,13 +1,16 @@
...
+Seeder.new(
+  FloodRiskEngine::Exemption,
+  'flood-risk-engine/exemptions.yml',
+  name_method: :code,
+  attributes_method: :summary
+).build
+
+puts Seeder.report
+
+
```
woodpigeon: Looks a bit nicer to me and doesn't lean on I18n which was maybe a bit confusing before. @autotelik this came out of trying to seed Heroku (env = production) so had to find a way to seed without just relying on Rails.env. Event in live env we will need to run some initial seeds to set up database. Checking presence of record first prevents it being added again. However I think I'd feel safer if we didn't attempt seeding Exemptions at all if any Exemptions exist?

db/seeds.rb
reggieb: No. At the moment there doesn't appear to be any requirement for it to be environment specific.

db/seeds.rb
reggieb: When I've used this approach before, it has been useful to allow seeds to update objects like this. For example, if the wording of an exemption needs to change, you can update the seed file to match the new wording, and it can be rolled out into production via seeding.

Alternatively, I can add `unless FloodRiskEngine::Exemption.exist?` to the end of line 12.

woodpigeon: On reflection I think you are right - this is a good way in this instance of getting exemption changes into a deployed env given that there is no other way other than firing up a db console.
If seeding other things (like users in the BO) I'd be more cautious maybe. But I think in the FO its only exemptions. :shipit:

PR #41, opened 2016-04-28
Ensure errors are rendered via a redirect to show.

reggieb:
Fix to issue https://github.com/EnvironmentAgency/flood-risk-engine/issues/38

It changes the update step action so that it redirects to show with the params that generated the error.


autotelik:
AFAIK based on ad hoc browser testing this fixes the issue with Back button after errors encountered

PR #43, opened 2016-04-29
Change show to use form method `validate` when checking for errors.

reggieb:
Calling form.valid? does not work in the way I thought. Need to pass the params to the form via the `validate` method to get the form to register the errors correctly.

### DIFF
--- a/app/validators/flood_risk_engine/name_format_validator.rb
+++ b/app/validators/flood_risk_engine/name_format_validator.rb
@@ -5,8 +5,8 @@ module FloodRiskEngine
   class NameFormatValidator < ActiveModel::EachValidator
 
     def validate_each(record, attribute, value)
-      if(value !~ NameFormatValidator.valid_name_regex)
-        record.errors.add attribute,  I18n.t("flood_risk_engine.validation_errors.name.invalid")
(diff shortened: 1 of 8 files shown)


app/validators/flood_risk_engine/name_format_validator.rb
```diff
@@ -5,8 +5,8 @@ module FloodRiskEngine
   class NameFormatValidator < ActiveModel::EachValidator
 
     def validate_each(record, attribute, value)
-      if(value !~ NameFormatValidator.valid_name_regex)
-        record.errors.add attribute,  I18n.t("flood_risk_engine.validation_errors.name.invalid")
+      if value !~ NameFormatValidator.valid_name_regex
```
autotelik: is this rubocop removing '('  from ifs ?

autotelik:
perhaps I haven't groked it correctly, but overall seems pretty inefficient scheme, if we redirect and rerun validations every time there is an error, rather than just rendering the view with the error

autotelik:
feature tests are now passing so this appears to fix lost validation messages - as mentioned over all scheme seems odd  but  :ship:

PR #44, opened 2016-04-29
RIP-73 - location name specs

autotelik:
Regression in code base was not caught, these try to address but suspect validations still not properly tested through whole stack.

Prefer bringing in proper feature specs which did catch the regression

### DIFF
--- a/config/locales/flood_risk_engine/enrollments/steps/show/generic.yml
+++ b/config/locales/flood_risk_engine/enrollments/steps/show/generic.yml
@@ -13,8 +13,6 @@ en:
           heading: Review
         user_type:
           heading: User Type
-        local_authority:
-          heading: Local Authority
         local_authority_address:
           heading: Local Authority Address
         main_contact_name:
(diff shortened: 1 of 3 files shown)


autotelik:
:eyes:

reggieb:
I think the `local_authority_spec.rb` should be at `spec/controllers/flood_risk_engine/enrollments/steps_controller/local_authority_spec.rb` as it is describing some of the behaviour of the steps controller.

reggieb:
Otherwise, looks good to me.

autotelik:
I don't think there are really testing the controller, which should be transparent to us, these are testing the enrollment gets built, validates etc hence location

reggieb:
The start of the spec is:
```
require "rails_helper"
module FloodRiskEngine
  describe Enrollments::StepsController, type: :controller do
```

autotelik:
develop/spec/controllers/flood_risk_engine/enrollments/steps_controller_spec.rb

PR #45, opened 2016-04-29
Update exemption seeds with latest descriptions

woodpigeon:
Update exemption descriptions, taken from the spreadsheet column
called _Title: GOV.UK PDF Sprint 4 agreed text_


woodpigeon:
Text changes only so self-merged.

PR #46, opened 2016-05-03
Feature/rip 72 business types page local auth or public body

reggieb:
The main aim RIP 72 is to add the pathway from choosing the user's organisation type as "Local authority or Public Body", to the page where the user adds the name of the organisation.

However, to make sure that works, I needed to make sure that in choosing that pathway, the user wasn't taken on one of the other pathways. So I've also added minimal place-holders for the other workflows, and ensured that the code used to complete this task also allows users to choose other organisation types.

So please note: That I expect the additions to the generic.yml locale file to be removed when the other journeys are worked on. That the workflows for the other journeys will include more than one step once work starts on those user journeys.


reggieb:
Rebased

app/state_machines/flood_risk_engine/enrollment_state_machine.rb
```diff
@@ -14,8 +14,54 @@ class EnrollmentStateMachine < FiniteMachine::Definition
...
+      )
+
+      event :go_forward, WorkFlow.for(:limited_company).merge(
+        if: -> { target.org_type == "limited_company" }
+      )
+      event :go_back, WorkFlow.for(:limited_company).invert.merge(
+        if: -> { target.org_type == "limited_company" }
+      )
+
+      event :go_forward, WorkFlow.for(:limited_liability_partnership).merge(
```
autotelik: couldn't this all go in some sort of loop, based on the complete list of org_type ?  .. make it much more readable and less brittle to change

reggieb: I've tried a number of ways of doing just that, but can't get any of them to work. The problem is that the `target` object is defined on the fly by the underlying state engine, and the whole script is defined within the code of the state engine's configuration. That means that you can just add loops or methods that would tidy this up.

app/forms/flood_risk_engine/steps/user_type_form.rb
```diff
@@ -21,7 +21,13 @@ def params_key
       # when we save the enrollment
       def save
         model.org_type = org_type.to_sym
+        super
+        # Need to go back and then forward as organisation change will effect
+        # onward journey. Without the back and forth a user cannot go back
+        # through the user type selection.
+        enrollment.go_back
```
autotelik: this state engine appear to be very complex to work with, state management appears to be  expanding out from being encapsulated in the state machine itself to  lots of places where it's used. i.e as a developer not clear what this back/fwd stuff and double saved achieves, or under what circumstances I might need to do something similar

app/state_machines/flood_risk_engine/enrollment_state_machine.rb
```diff
@@ -14,8 +14,54 @@ class EnrollmentStateMachine < FiniteMachine::Definition
...
+
+      event :go_forward, WorkFlow.for(:other).merge(
+        if: -> { target.org_type == "other" }
+      )
+      event :go_back, WorkFlow.for(:other).invert.merge(
+        if: -> { target.org_type == "other" }
+      )
     end
   end
 end
```
woodpigeon: I have aboslutely no idea if this runs, but treat as a mad pseudo code idea for reducing noise a bit

```ruby
events do
  %i(partnership, other, individual).each do |organisation_type|
    is_current = -> { target.organisation_type == organisation_type.to_s }
    create_go_forward_event_for(organisation_type, is_current)
    create_go_back_event_for(organisation_type, is_current)
  end
end

def self.create_go_forward_event_for(organisation_type, org_type_is_current)
  steps = WorkFlow.for(organisation_type)
  event :go_forward, steps.merge(if: org_type_is_current))
end

def self.create_go_back_event_for(organisation_type, org_type_is_current)
  steps = WorkFlow.for(organisation_type).invert
  event :go_back, steps.merge(if: org_type_is_current)
end
```

reggieb: You get: 

```
undefined method `create_go_forward_event_for' for #<FiniteMachine::EventsDSL:0x00000003de82c8>
```
Same if you make them instance methods. As I said, I have tried that sort of thing already.

reggieb: However, if you move the method to the EventsDSL class, you get close:

```
require "finite_machine"
class FiniteMachine::EventsDSL
  def create_go_forward_event_for(organisation_type, org_type_is_current)
    steps = FloodRiskEngine::WorkFlow.for(organisation_type)
    event :go_forward, steps.merge(if: org_type_is_current)
  end

  def create_go_back_event_for(organisation_type, org_type_is_current)
    steps = FloodRiskEngine::WorkFlow.for(organisation_type).invert
    event :go_back, steps.merge(if: org_type_is_current)
  end
end
```

reggieb:
The refactor of the state machine has lost the original in-line comments. I think the result is much cleaner.

app/forms/flood_risk_engine/steps/user_type_form.rb
reggieb: I agree that this isn't perfect, but was the simplest way I could get the backward functionality to work. This back and forth should only be needed if a attribute change alters which work flow is being used. So at the moment - only when the organisation type changes.

BTW, `super` is saving the `organisation` object and not the enrollment.

reggieb:
Rebased again

autotelik:
:ship: from tom

reggieb:
Rebased again

PR #48, opened 2016-05-03
Implement Contact for the activity page (Correspondence Contact) - RIP-75

autotelik:
https://eaflood.atlassian.net/browse/RIP-75

Add ReForm class and build validations
Add view partial based on prototype
Extract text and store in  locale data
Add new Validator that can check number of words entered


app/forms/flood_risk_engine/steps/correspondence_contact_name_form.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      def self.factory(enrollment)
+        contact = enrollment.correspondence_contact
+
+        unless contact
+          contact = enrollment.correspondence_contact = FloodRiskEngine::Contact.new(contact_type: :correspondence)
+        end
+
+        new(contact, enrollment)
+      end
```
woodpigeon: Could poss do #initialize in 2 lines without losing readability
```
def self.factory(enrollment)
   enrollment.correspondence_contact ||= FloodRiskEngine::Contact.new(contact_type: :correspondence)
   new(enrollment.correspondence_contact, enrollment)
end
```

app/forms/flood_risk_engine/steps/correspondence_contact_name_form.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      validates :position, 'flood_risk_engine/text_field_content': true, allow_blank: true
+
+      validates :position, length:
+        {
+          maximum: CorrespondenceContactNameForm.position_max_length,
+          message: I18n.t("#{CorrespondenceContactNameForm.locale_key}.errors.position.too_long",
+                          max_length: CorrespondenceContactNameForm.position_max_length)
+        }
+
```
woodpigeon: Alternative layout, not necessarily better
```ruby
property :full_name
property :position

validates :full_name, 
  presence: {
    message: I18n.t("#{self.class.locale_key}.errors.full_name.blank")
  },
  'flood_risk_engine/text_field_content': {
    allow_blank: true
  },
  length: {
    maximum: self.class.name_max_length,
    message: I18n.t("#{self.class.locale_key}.errors.full_name.too_long",
                    max_length: self.class.name_max_length)
  }

validates :position,
  'flood_risk_engine/text_field_content': {
     allow_blank: true
  },
  length: {
    maximum: self.class.position_max_length,
    message: I18n.t("#{self.class.locale_key}.errors.position.too_long",
                    max_length: self.position_max_length)
  }
```

app/forms/flood_risk_engine/steps/correspondence_contact_name_form.rb
```diff
@@ -0,0 +1,78 @@
...
+          contact = enrollment.correspondence_contact = FloodRiskEngine::Contact.new(contact_type: :correspondence)
+        end
+
+        new(contact, enrollment)
+      end
+
+      def self.locale_key
+        "flood_risk_engine.enrollments.steps.correspondence_contact_name"
+      end
+
```
woodpigeon: would it help to move locale_key to the base form? eg as instance method
```ruby
def locale_key
  "flood_risk_engine.enrollments.steps.#{params_key}"
end
```

app/forms/flood_risk_engine/steps/correspondence_contact_name_form.rb
autotelik: yes agreed. much better.

app/forms/flood_risk_engine/steps/correspondence_contact_name_form.rb
autotelik: I don;t like the validaitons tlye myself, find that hard to differnentiate the individual validators

On grouping, I'm torn, usually I keep thigns togther as per your example,  e.g properties but then when their usage blocks starts to get too long, my instinct is to move the variable declrations next to the usage .. but happy to re group them,if thats a standard we're adopting, at top, as per 

property :full_name
property :position

app/forms/flood_risk_engine/steps/correspondence_contact_name_form.rb
autotelik: I don't think an instance is needed to call these methods

 Things is you need an enrollment to construct this class so if you want to use this information in situations - like interrogating locale information, where you don't have or need an enrollment - you can't.

    let(:locale_errors_key) { "#{CorrespondenceContactNameForm.locale_key}.errors" }

 So actually I've been wishing param key was a class method, as like locale key really its class level info, rather than instance level. 

But I'm guessing theirs issues with inheritance chain if params_key was a class method. AFAIK it's  permissible to have both so maybe we can have  
```
      def params_key
        self.class.params_key
      end

      def self.params_key
        :correspondence_contact_name
      end
```

woodpigeon: Yeah I like that

app/forms/flood_risk_engine/steps/correspondence_contact_name_form.rb
woodpigeon: I personally like the grouped properties if they are the top of the class as you can see at a glance what class is about. I'm easy either way on validation style though.

woodpigeon:
:ship: once you are happy with your params key change (if u decide to do it) and maybe the grouped properties

app/forms/flood_risk_engine/steps/correspondence_contact_name_form.rb
autotelik: done.

autotelik:
@woodpigeon you might want to quickly re review what I've done with the new module .. all specs pass including full features for the 3 pages. Tried to standardise format of form with groups .. class methods, properties, validations

PR #51, opened 2016-05-04
Use PostgreSQL in test dummy app.

reggieb:
With this change, the dummy test environment will use the same database engine as the front and back end apps.


reggieb:
Rebased

spec/dummy/.env.example
```diff
@@ -0,0 +1,7 @@
+# This project uses environment variables for configuration
+# Copy .env.example and rename as .env in the project root
+# and add correct values. You can also use an environment specific file
+# e.g. .env.development
+
+DB_USER = "flood_risk"
```
autotelik: worth keeping things consistent with FO ?

```
## Database settings
PG_HOST='localhost'
PG_DATABASE='database-name'
PG_USERNAME='username'
PG_PASSWORD='password'

PG_TEST_DATABASE=database-name
PG_TEST_USERNAME=username
PG_TEST_PASSWORD=password
```

spec/dummy/config/database.yml
```diff
@@ -5,21 +5,26 @@
 #   gem 'sqlite3'
 #
 default: &default
-  adapter: sqlite3
+  adapter: postgresql
+  encoding: unicode
   pool: 5
-  timeout: 5000
+  host: localhost
```
autotelik: as above .. worth keeping things consistent with FO ?  .. just means can copy and paste from existing .env

spec/dummy/.env.example
reggieb: Yes - Using the PG_ prefix rather than DB makes sense. 

However, I don't think we should add environment variables that aren't used.

spec/dummy/config/database.yml
reggieb: Does anyone use a non-localhost database when running locally?

The travis configuration is handled via the database.travis.yml.

reggieb:
Rebased

PR #52, opened 2016-05-04
Update organisation type labels via locale, to match prototype

reggieb:
Reviewing the results of the RIP 72 update in the front end app: the organisation name labels need to be updated.


autotelik:
:ship:

PR #56, opened 2016-05-05
Move error params from url to session store

reggieb:
Store the error data in a session key rather than the redirect url, as data was exposed, and could be too long for redirect url.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,7 @@ PATH
   remote: .
   specs:
     flood_risk_engine (0.0.1)
+      activerecord-session_store (~> 0.1)
       dibber (~> 0.5)
       dotenv-rails (~> 2.1)
       finite_machine (~> 0.10)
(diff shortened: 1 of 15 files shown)


reggieb:
This is working, but is storing the session in the cookie space rather than in the database. I've put initializers in both the engine and the dummy app, but to no effect.

woodpigeon:
Remove `spec/dummy/config/initializers/session_store.rb` as that initializer is loading after yours

reggieb:
Of course!

woodpigeon:
Noticed a subtle navigation error (may not be part of this PR but can't reproduce in current develop)  
- go as far as 'Who should we contact about this activity?' without invokving any validation errors
- hit the back ​_link_​ twice to get back to 'What's the name of the local authority or public body?'
- now hit the back ​_button_​ and you can't go back - wants to stay on that page

autotelik:
@woodpigeon I'm seeing weird navigation related behaviour too, Back button can disappear, and the State can becoem stuck in one state (Continue doesn't move the page on) ..  not 100% but my guess would be it's related to the Organisation association - seems like the aspects of the navigation that relies on OrgType can stop working - e.g when no organisation present on Enrollment -  but does not error or throw an exception

reggieb:
How are you getting to the 'Who should we contact about this activity?' step? Are you testing this via the dummy app, or are you hosting it in the front-office app?

reggieb:
I get the same problem if I use the current `develop` branch. However, on the code used in this branch, the problem does not arise.

woodpigeon:
:ship: it

reggieb:
Actually the problem does still arise - it is the combination of both using the back link and the back button that causes the problem. If you click on the back link, and then the browser back button the browser will try to take you to the previous page - which is the one where you clicked on the back link - not the previous step!. An issue is being raise to address this separately.

PR #58, opened 2016-05-09
Rip 118 contact telephone number

reggieb:
This fixes the problems with the previous pull request for this feature.

Original request: https://eaflood.atlassian.net/browse/RIP-118


autotelik:
:ship - for the bigger picture I think we need more extensive tests - view/feature specs  - sine the form specs are ok for testing validations but are not catching issues across the full stack

PR #66, opened 2016-05-10
Implement Postcode Lookup for LocalAuthority - RIP-74 

autotelik:
https://eaflood.atlassian.net/browse/RIP-74

Due to nature of coding this ticket, this PR contains  code for two tickets, but does not complete to a QA standard. This PR forms a basis for incorporating Manual entry into the complete flow, and will enable completion in separate PR of all edge cases and validation scenarios.

Further work will also include moving address related actions to new AddressController (coded in separate manual address entry ticket)

Specific Tasks : 
    https://eaflood.atlassian.net/browse/RIP-134
    https://eaflood.atlassian.net/browse/RIP-135

* Create Form and View
* Include ea-address_lookup gem
* Configure in an initializer with test address facade url etc and a 5 second timeout
* Validate post code
* Implement redirect of postcode to post code lookup action.
* Implement action to call address page with valid postcode
* Selected address is saved against Organisation
* Ensure basic  Back button/link navigation works to populate address drop down
* Implement 'Change' link back to Postcode selection page

The scenarios covered have been copied from RIP-74 into RIP-134 and RIP-135

Full Feature tests for these pages can be found here : 

https://github.com/EnvironmentAgency/flood-risk-front-office/pull/12/files


autotelik:
:eyes:

woodpigeon:
@autotelik Have some comments but have not added yet - I tried running locally and get `translation missing: en.flood_risk_engine.enrollments.steps.local_authority_address.num_addresses_found` after submitting postcode step - so is it worth waiting until the process_postcode and select address stuff is ready before reviewing the whole piece?

autotelik:
why could this not be merged  ? if its just that locale key, I'll add now.

I was trying to keep the PR manageable so thought it easier to get this in - covers the postcode page RIP-134 -  with partial implementation of address - enough to  enable the journey to be continued, pretty much like Ron's generic.yaml and null forms

autotelik:
There are also feature specs for the post code page here

https://github.com/EnvironmentAgency/flood-risk-front-office/pull/12

All other feature specs pass too, so I don't believe there is any regression

autotelik:
@woodpigeon  p.s if you want me to do all the work on this one PR I'm more than happy to

woodpigeon:
@autotelik Yes I'd say add the next page (select address) as part of this PR if you are happy with that? -  it might be easier for @reggieb then to integrate the manual address page if he can see in one PR how you are building out the lookup process. There might be some changes that fall out of the comments on this PR as its quite a big one, but we can catch up on Monday and iron anything out

autotelik:
:eyes:    This code is QA ready for post code search page i.e :local_authority_postcode  but not the following address page which requires further integration with other PRs

.travis.yml
```diff
@@ -1,43 +1,36 @@
...
-# PR's and pushes. However this means when we push to a PR we have to wait for Travis to finish
-# 2 builds. If we unticked 'pushes' when the PR was finally merged that would not be built. The
-# brute force approach would be to untick build PR's and just build all pushes. We instead have
-# gone with the approach outlined here http://stackoverflow.com/a/31882307
 branches:
   only:
-    - master
-    - develop
+  - master
+  - develop
```
woodpigeon: Looks like the yaml array formatting has lost its indentation throughout the file - its getting parsed ok but I'd vote for restoring the indetation

autotelik: interestingly the travis cmd line tool did this when generating new secure env vars

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -8,30 +8,81 @@
...
-          redirect_to(form.redirection_url)
-        elsif success
-          step_forward
-          redirect_to step_url
+
+        # The redirect was happening regardless of outcome, hence moved to within success
+        # i.e if validate or save fails we should NOT redirect
+        if success
+
+          if form.redirect?
```
reggieb: Redirecting only after successful validation breaks the redirect behaviour in previous steps where a redirection was used to exit the process under certain situations. Instead the form should set `form.redirect?` to false if redirection is to be prevented.

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -8,30 +8,81 @@
...
+            redirect_to step_url
+          end
+
         else
           handle_failure_and_redirect_back_to_show
         end
       end
 
+      def process_postcode # rubocop:disable Metrics/AbcSize
+        # We redirect here from update when dealing with page containing Postcode entry and form validated
```
reggieb: Why? Why are we treating `postcode` differently to every other attribute. 

Compare Contact and Address. For a contact we apply the telephone number in one step, and the email address in another. So why don't we apply an address postcode in one step, and the rest of the address in another.

This redirection is just adding unnecessary complication to the process.

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -8,30 +8,81 @@
...
+          form: address_form,
+          step: enrollment.step
+        }
+
+        logger.info("Rendering step #{enrollment.step} for Postcode [#{params[:post_code]}]")
+
+        render :show, locals: locals_after_redirect
+      end
+
+      def process_address
```
reggieb: Again - why is address being handled differently here?

app/factories/flood_risk_engine/steps/form_object_factory.rb
```diff
@@ -42,6 +42,7 @@ def setup_form_object(step)
           begin
             form_name.constantize
           rescue NameError => x
```
reggieb: Why are we rescuing the NameError error here? If we let it bubble to the surface it will appear in a stand log output.

app/forms/flood_risk_engine/steps/base_address_form.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      validates :uprn, numericality: {
+        only_integer: true,
+        allow_blank: true,
+        message: I18n.t("flood_risk_engine.validation_errors.uprn.format")
+      }
+
+      # Post code is set in previous page, and must be passed in so we can create the
+      # address drop down list - need this externally set as we can't access Session in the Form
+      attr_reader :post_code
```
reggieb: To be consistent with the database attribute this should be `postcode`

app/forms/flood_risk_engine/steps/base_address_form.rb
```diff
@@ -0,0 +1,131 @@
...
+        # We assume postcode validation already occured so do not validate the format here
+        @post_code = options[:post_code] if options.key? :post_code
+      end
+
+      # This is a form reader - provides view with addresses for drop down selection
+      def address_list
+        @address_list ||= find_by_postcode
+      end
+
+      # The normal update action is no good for managing addresses, need to wire up continue button to use
```
reggieb: "The normal update action is no good for managing addresses" Why?

app/forms/flood_risk_engine/steps/base_address_form.rb
```diff
@@ -0,0 +1,131 @@
...
+      def assign_via_uprn
+        # Use UPRN to find address data
+        address_attributes = find_by_uprn
+
+        if address_attributes
+          # Build and validate the inbound address data against a real Flood Address model
+          logger.debug("Build LocalAuthority address from #{address_attributes.inspect}")
+          @address_to_assign = Address.new(address_attributes)
+
+          return @address_to_assign.validate
```
reggieb: Calling `validate` on an Address instanced will run the validations on the model. I assumed by using forms, we had made the decision that those forms controlled validation. If we start using the model validation as well as form validation, I think we will get in a mess.

app/forms/flood_risk_engine/steps/base_address_form.rb
```diff
@@ -0,0 +1,131 @@
...
+        end
+
+        false
+      end
+
+      def find_by_uprn
+        logger.debug("Start BaseAddressForm Lookup on UPRN #{uprn}")
+
+        finder = FloodRiskEngine::FindAddressByUprn.new(uprn)
+        inbound = finder.search
```
reggieb: I'd suggest tidying this up by adding a class method to FloodRiskEngine::FindAddressByUprn that initiated a new instance and called its search method. Something like:
```ruby
def self.search(args={})
  new(args).search
end
```
Then here you can just:
```
inbound = FloodRiskEngine::FindAddressByUprn.search(uprn)
```

app/forms/flood_risk_engine/steps/base_address_form.rb
```diff
@@ -0,0 +1,131 @@
...
+
+        # parser used to convert inbound data into a format for use in Views
+        parser = FloodRiskEngine::Services::Address::Deserialize::EaFacadeToAddress.new
+
+        parser.selectable_address_data(inbound)
+      end
+
+      def initialize(model, enrollment)
+        @post_code = ""
+        @address_to_assign = nil
```
reggieb: You don't need to assign instance variables as nil on initialization as they are already nil.

app/forms/flood_risk_engine/steps/base_form.rb
```diff
@@ -19,6 +19,12 @@ def redirect?
         false
       end
 
+      # Default to the url as defined in the view/partial but provide opportunity for Forms to
+      # wire up the continue button as they see require
+      def url
```
reggieb: I'm not sure being able to change the continue button is necessary or desirable, but if it is the name should be meaningful. For example: `continue_button_url`

app/forms/flood_risk_engine/steps/base_postcode_form.rb
```diff
@@ -0,0 +1,20 @@
+module FloodRiskEngine
+  module Steps
+    class BasePostcodeForm < BaseRedirectableForm
+
+      property :post_code, virtual: true
```
reggieb: The property should be `postcode`

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
autotelik: Because the designers want this selection over two pages

Its not an attribute at this stage. Its a request for data. I am not storing half baked data that will add unnecessary complication in its management. I am storing the full address once they select that full address.

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
autotelik: The address requires an external service, It is not like the other forms where you simply take the data validate  and store.

app/factories/flood_risk_engine/steps/form_object_factory.rb
autotelik: you told me to in a previous PR

app/forms/flood_risk_engine/steps/base_address_form.rb
autotelik: this is not a database attribute

app/forms/flood_risk_engine/steps/base_address_form.rb
autotelik: I need to set it to my own actions, process_address and process_postcode since they need to do quite different things to update

app/forms/flood_risk_engine/steps/local_authority_address_form.rb
```diff
@@ -0,0 +1,49 @@
...
+
+        new(FloodRiskEngine::Address.new(address_type: :primary), enrollment)
+      end
+
+      def prepopulator(options = {})
+        super(options)
+
+        # This is first stab at ensuring we can handle Back button/link
+        if @post_code.blank?
+          existing_address = enrollment.organisation.primary_address
```
reggieb: I'd suggest in `enrollment` delegating `primary_address` to `organisation` as this would greatly simplify this bit of code - that it, you wouldn't need to do so many checks that objects in the chain exist.

app/forms/flood_risk_engine/steps/base_address_form.rb
autotelik: this is validating data from an external service,the code for  RIP-135 is not complete, this is more a place holder for the concept - that validate that address facade data can be used to build Address

so agreed, once your manaul address validation stuff is in validators I can use that

app/models/flood_risk_engine/organisation.rb
```diff
@@ -12,5 +12,8 @@ class Organisation < ActiveRecord::Base
       other: 5,
       unknown: 6
     }
+
+    belongs_to :primary_address, class_name: "FloodRiskEngine::Address"
```
reggieb: This should be a `has_one`:
```ruby
has_one :primary_address, class_name: "FloodRiskEngine::Address", as: :addressable
```

app/services/flood_risk_engine/find_address_by_postcode.rb
```diff
@@ -0,0 +1,31 @@
...
+    def initialize(post_code)
+      Rails.logger.debug("FindAddressByPostcode supplied with params [#{post_code.inspect}]")
+      @post_code = post_code
+      @address_result_set = []
+    end
+
+    def search
+      @address_result_set = []
+
+      inbound = begin
```
reggieb: `inbound` is redundant. This would give the same results:
```ruby
def search
   @address_result_set = []
   EA::AddressLookup.find_by_postcode(post_code)
rescue => ex
   Airbrake.notify(ex) if defined? Airbrake
   Rails.logger.error "Address lookup service failed: #{ex}"
   []
end
```

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
autotelik: there were no regressions and seemed cleaner but reverted.

app/services/flood_risk_engine/find_address_by_uprn.rb
```diff
@@ -0,0 +1,31 @@
+module FloodRiskEngine
+  class FindAddressByUprn
+
+    attr_reader :address_data
+
+    def initialize(uprn)
+      Rails.logger.debug("FindAddressByuPRN supplied with params [#{uprn.inspect}]")
+      @uprn = uprn
+      @address_data = ""
```
reggieb: I think there needs to be a consistent way of clearing the @address_data. Either `@address_data = ""` or `@address_data = nil`

app/views/flood_risk_engine/enrollments/steps/_local_authority_address.html.erb
```diff
@@ -1 +1,34 @@
...
+  %>
+</p>
+
+<%= f.hidden_field :post_code, value: form.post_code %>
+
+<label class="form-label" for="address_match_selection">
+  <%= t('.select_from_list_label') %>
+</label>
+
+<% form.address_list.unshift({ moniker: t('.num_addresses_found', num:  form.address_list.count) }) %>
```
reggieb: This should be part of the list building process, and is logic that is too easily lost within the view

app/services/flood_risk_engine/find_address_by_postcode.rb
autotelik: agreed

app/services/flood_risk_engine/find_address_by_uprn.rb
autotelik: agreed. fixed

app/views/flood_risk_engine/enrollments/steps/_local_authority_address.html.erb
reggieb: The list count should not be added to the list in this way - use the select form helper's prompt option

app/views/flood_risk_engine/enrollments/steps/_local_authority_address.html.erb
```diff
@@ -1 +1,34 @@
...
+
+<fieldset>
+  <%= form_group_and_validation(form, :uprn) do %>
+    <%= f.select :uprn,  options_for_select( form.address_list.collect { |a| [ a[:moniker], a[:uprn] ] }), {},
+                 { include_blank: false,
```
reggieb: Instead of `include_blank: false` use:
```
prompt: t('.num_addresses_found', num:  form.address_list.count)
```
Then you don't have to add the list count to the list of addresses, and you can let the select form helper handle what to set the value as.

config/initializers/address_lookup.rb
```diff
@@ -0,0 +1,8 @@
+EA::AddressLookup.configure do |config|
+  config.address_facade_server  = ENV["ADDRESS_FACADE_TEST_SERVER"]
```
reggieb: I think we should only be using a test server, when in the test environment - or perhaps in develop in some circumstances. So the environment variable should be `ENV["ADDRESS_FACADE_SERVER"]`

config/initializers/address_lookup.rb
```diff
@@ -0,0 +1,8 @@
...
+  config.address_facade_port    = ENV["ADDRESS_FACADE_TEST_PORT"]
```
reggieb: `ENV["ADDRESS_FACADE_PORT"]`

db/migrate/20160510174948_add_primary_address_organisation.rb
```diff
@@ -0,0 +1,10 @@
+class AddPrimaryAddressOrganisation < ActiveRecord::Migration
```
reggieb: This migration is unnecessary - there is already a mechanism for attaching address to objects

app/models/flood_risk_engine/organisation.rb
autotelik: Ok given its the primary address I added migration so organisation had direct primary_address_id  

I'll remove migrations and change to has_one

lib/flood_risk_engine/engine.rb
```diff
@@ -6,6 +6,9 @@ module FloodRiskEngine
   class Engine < ::Rails::Engine
     isolate_namespace FloodRiskEngine
 
+    # Add a load path for this specific Engine
+    config.autoload_paths += Dir["#{config.root}/lib/**/"]
```
reggieb: I'm not sure having some code in `lib` and some in `app` is a good idea. Why not just keep everything together in `app`

lib/flood_risk_engine/services/address/deserialize/ea_facade_to_address.rb
```diff
@@ -0,0 +1,81 @@
...
+
+          # Builds collection of attributes in correct format to simply
+          # pass to Address new or create - to build an Address (plus Location)
+
+          def address_data(inbound) # rubocop:disable Metrics/MethodLength:
+            addresses = inbound["results"] if inbound
+
+            addresses ||= []
+
+            results = addresses.collect do |a|
```
reggieb: I'm not sure this doing anything meaningful except setting organisation to `""` if absent. Addresses can be built with symbol or hash keys.

lib/flood_risk_engine/services/address/deserialize/ea_facade_to_address.rb
```diff
@@ -0,0 +1,81 @@
...
+              {
+                easting: EaFacadeToAddress.format_xy(a["x"]),
+                northing: EaFacadeToAddress.format_xy(a["y"])
+              }
+            end
+
+            results
+          end
+
+          def self.format_xy(inbound)
```
reggieb: Can `format_xy` be an instance method? It would simplify the `location_data` method if it was.

spec/dummy/config/initializers/address_lookup.rb
```diff
@@ -0,0 +1,8 @@
+EA::AddressLookup.configure do |config|
+  config.address_facade_server  = ENV["ADDRESS_FACADE_TEST_SERVER"]
+  config.address_facade_port    = ENV["ADDRESS_FACADE_TEST_PORT"]
```
reggieb: `ENV["ADDRESS_FACADE_SERVER"]` and `ENV["ADDRESS_FACADE_PORT"]`

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -32,6 +36,46 @@ def update
...
       end
 
+      def process_postcode # rubocop:disable Metrics/AbcSize
+        # We redirect here from update when dealing with page containing Postcode entry and form validated
+        enrollment.go_forward
+        enrollment.save
+
+        # form & step cache results so don't return what we need after go_forward - so this ensures the right values
+        address_form = Steps::FormObjectFactory.form_object_for(enrollment.step, enrollment)
+
```
woodpigeon: If you break out address_form to a method, you can remove the # rubocop:disable Metrics/AbcSize
```
def address_form
  @address_form ||= Steps::FormObjectFactory.form_object_for(enrollment.step, enrollment)
end
```

app/forms/flood_risk_engine/steps/base_address_form.rb
autotelik: I prefer that you can rerun the search without instantiating uprn.

app/forms/flood_risk_engine/steps/base_form.rb
autotelik: This is the url to use in the `form_for` .. so in view this reads as `form.url` .. not for use by the continue button per se

app/forms/flood_risk_engine/steps/base_postcode_form.rb
autotelik: its virtual and purposelessly named as to be different to address. this form is not backed by a model

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
reggieb: It's data that the user has input and that needs to be stored. 

The unnecessary complication of storing the postcode and moving on to add the other address attributes seems a lot less complicated than the necessary complication you've added so as to be able to handle this in a special way.

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
autotelik: ok

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
reggieb: A form can use a service object to gather data without this special code.

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -32,6 +36,46 @@ def update
...
+      def process_address
+        # we need to prepopulate this form, in case of errors that mean address list needs repopulating
+        form.prepopulator(params) if form.respond_to?(:prepopulator)
+
+        if form.validate(params)
+          form.save
+          step_forward
+          redirect_to(step_url) && return
+        else
```
woodpigeon: I don't think you need the `return` here?

app/factories/flood_risk_engine/steps/form_object_factory.rb
reggieb: If that is the case, I was wrong.

app/forms/flood_risk_engine/steps/base_address_form.rb
reggieb: But the data it holds will be eventually.

app/forms/flood_risk_engine/steps/base_address_form.rb
reggieb: There's nothing to stop you doing that when necessary. It doesn't seem to be necessary within this method.

app/forms/flood_risk_engine/steps/base_form.rb
reggieb: OK

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
autotelik: btw this does highlight that the caching used on form and step is very dangerous in a state machine

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
autotelik: I thought it was good practice ?

app/views/flood_risk_engine/enrollments/steps/_local_authority_address.html.erb
autotelik: ok

spec/dummy/config/initializers/address_lookup.rb
autotelik: yep agreed. moved to secrets

app/factories/flood_risk_engine/steps/form_object_factory.rb
woodpigeon: A couple of times during dev I have had to comment out the error handler in `setup_form_object` as it swallows up the underlying error. I'd be happier to lose it altogether if possible. If we are going to resolve the form class both dynamically and via a hash map (which we are only doing because we cannot agree :smile:) then how about we delete `setup_form_object` and change `form_object_for` to
```
def form_object_for(step, enrollment)
  klass = form_object_class_map[step.to_sym]
  klass ||= "FloodRiskEngine::Steps::#{step.to_s.classify}Form". constantize
  klass.factory(enrollment)
end
```
Its just as clear what is happening (though a future developer is still going to be scratching their heads as to why we are trying both lookup methods!), and the error raised if the form does not exists will be `#<NameError: uninitialized constant FloodRiskEngine::Steps::BlahStepNonsenseForm>` which is all a developer needs to see what's happening.

app/forms/flood_risk_engine/steps/base_address_form.rb
woodpigeon: is it `post_code` to differentiate it from `postcode` in the db?

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
autotelik: I can't recall all the issues I had with your generic methods now , but I tried alot of things, and this  was the way I got everything working and clean behaviour for back buttons .

When the manual entry element is merged I can revisit these.

config/initializers/address_lookup.rb
autotelik: agreed. switched to secrets

app/forms/flood_risk_engine/steps/base_address_form.rb
```diff
@@ -0,0 +1,131 @@
...
+        inbound = finder.search
+
+        # parser used to convert inbound data into a format we can
+        # use to construct an Address
+        parser = FloodRiskEngine::Services::Address::Deserialize::EaFacadeToAddress.new
+
+        flood_address_data = parser.address_data(inbound)
+
+        logger.info("Parsed address #{flood_address_data.first}")
+
```
woodpigeon: Any `logger.info` calls will be in the production log by default - is that the intention or is this just for visibility in dev/test? I noticed in WEX production logs there is quite a bit of logging noise (some produced by me) and wonder if we should dial things down a bit in RP and avoid `info` unless really useful for live debugging (which is hard to predict I know)? In WEX, have we had to inspect the logs at any point so far, and been grateful for the extra logging output? If not, it might suggest we don't need to log much. Any thoughts?

app/factories/flood_risk_engine/steps/form_object_factory.rb
autotelik: if you wanna just use your hash map go for it

app/forms/flood_risk_engine/steps/base_address_form.rb
autotelik: Yes.  IMHO until they choose the address the association remains empty so its clear no address has been selected. I do not want to create an address with only a postcode, firstly that is simply not the address, then we have  abondonment, back button, back link, change postcode aand check details to handle .. and we are suddenly managing weird edge cases

app/forms/flood_risk_engine/steps/base_postcode_form.rb
woodpigeon: I see @reggieb's point though. Personally I'd call it `postcode` and have done with it ;-)
`post_code` as opposed to `postcode` doesn't tell a future develop much but could lead to confusion and introduction of a bug even. As an alternative what about `tmp_postcode` which has a bit more meaning?

app/forms/flood_risk_engine/steps/base_address_form.rb
autotelik: >>But the data it holds will be eventually.

Not necessarily. In every other form data is saved only after validation. Why should this be any different

app/forms/flood_risk_engine/steps/local_authority_address_form.rb
```diff
@@ -0,0 +1,49 @@
...
+
+          @post_code = existing_address.postcode if existing_address && existing_address.postcode.present?
+
```
woodpigeon: Alternative ```@post_code = existing_address && existing_address.postcode```

app/forms/flood_risk_engine/steps/base_address_form.rb
autotelik: dev/test .. happy to do a sweep after everything settles down and remove all my logging calls .... but having had to do some support for big ecommerce sites, my experience is Rails devs do not log enough, and  yes the wex logs were useful on a couple of the early bugs, although probably too much address related calls and data being logged ...

app/services/flood_risk_engine/find_address_by_postcode.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def search
+      @address_result_set = begin
+        EA::AddressLookup.find_by_postcode(post_code)
+      rescue => ex
+        Airbrake.notify(ex) if defined? Airbrake
+        Rails.logger.error "Address lookup service failed: #{ex}"
+        []
+      end
+
```
woodpigeon: I can see why the error handling is like this. Don't have a suggestion for this but in the rspec suite if a postcode test fails (eg VCR cassette url mismatch) the error does not show up and you have to dig in log/test.log to find the problem. Its the sort of thing a future developer could burn a couple of hours tracking down if they get a failing test like this. Maybe its poss to re-raise the error if `Rails.env.test?`and the error summary contains /VCR/?

app/services/flood_risk_engine/find_address_by_uprn.rb
```diff
@@ -0,0 +1,30 @@
...
+      @address_data = nil
+
+      @address_data = begin
+        EA::AddressLookup.find_by_uprn(uprn)
+      rescue => ex
+        Airbrake.notify(ex) if defined? Airbrake
+        Rails.logger.error "Address lookup service failed: #{ex}"
+        nil
+      end
+
```
woodpigeon: I might be wrong (sorry if so) but I don't think `@address_data` or the `address_data` are used anywhere so search could be
```
def search
  EA::AddressLookup.find_by_uprn(uprn)
rescue => ex
  Airbrake.notify(ex) if defined? Airbrake
  Rails.logger.error "Address lookup service failed: #{ex}"
  nil
end
```

app/forms/flood_risk_engine/steps/base_postcode_form.rb
autotelik: this form has nothing to do with the address model, it's a virtual attribute that relates to the search parameter and `post_code` matches how params defined in the route. 

This form relates to collecting  a group of addresses, only one of which will ever be persisted 

 for sake of moving on I'll rename it postcode

config/initializers/address_lookup.rb
```diff
@@ -0,0 +1,8 @@
...
+  config.address_facade_url     = "/address-service/v1/addresses/postcode"
+
+  config.address_facade_client_id =  ENV["ADDRESS_FACADE_CLIENT_ID"]
+  config.address_facade_key       =  ENV["ADDRESS_FACADE_KEY"]
+end
```
woodpigeon: would you mind if added explicitly the timeout setting too, so its clear its possible to change it where to change it if needed - its something that they might want to tweak
```
config.timeout_in_seconds = 5
```

app/forms/flood_risk_engine/steps/base_postcode_form.rb
autotelik: I take it back. not changing it. breaks too many specs

lib/flood_risk_engine/services/address/deserialize/deserialize_base.rb
```diff
@@ -0,0 +1,19 @@
...
+
+          def address_data(_inbound)
+            raise NotImplementedError, "No address_data implementation provided"
+          end
+
+        end
+      end
+    end
+  end
+end
```
woodpigeon: I could never see the point of this class in WEX ;-) (there's only one ever one sublcass) - I'd delete it if it were me

app/forms/flood_risk_engine/steps/local_authority_address_form.rb
autotelik: that form is not clear as to what  are you assigning, at first glance looks like @post_code is a boolean

app/services/flood_risk_engine/find_address_by_postcode.rb
autotelik: ok I'll look into that. I think originally the rescue had to be there to report properly to Airbrake

app/forms/flood_risk_engine/steps/local_authority_address_form.rb
autotelik: no thanks. its not the primary address of the enrollment

lib/flood_risk_engine/engine.rb
autotelik: Pattern: Store code that is not domain specific in lib

app/factories/flood_risk_engine/steps/form_object_factory.rb
autotelik: while we're on about time lost an d confusign developers, can we please remove that generic.yaml crap .. that's not production data and has lost me hours baffled why locale data was showing up wierd

reggieb: We need holding pages that render when one gets to a step that hasn't been built yet. Otherwise we'd have to build the steps in sequence and we'd not be able to test completion of a page (as success would lead to rendering of the next page which hadn't been built).

Without the generic locale file, the holding pages fail due to them not having heading locales defined.

We can get rid of the generic locale file when we no longer need holding pages.

autotelik: missing translation does not cause failure - they just show 'Heading'

app/forms/flood_risk_engine/steps/base_redirectable_form.rb
```diff
@@ -0,0 +1,18 @@
+module FloodRiskEngine
+  module Steps
+    class BaseRedirectableForm < BaseForm
```
reggieb: I can see where this class is used. Can it be removed?

autotelik: The PR is not ready to be reviewed yet. I moved the label as I work on all the requested changes.

But No I won't remove it, as this was useful,and may prove useful

app/models/flood_risk_engine/address_search.rb
```diff
@@ -0,0 +1,5 @@
+module FloodRiskEngine
+  class AddressSearch < ActiveRecord::Base
```
reggieb: Saving the postcode to the database has allowed the process to work without custom controller actions, which I think is good. Personally, I'd store the postcode on the address object being built rather than creating a special class to hold the postcode during address building, but I guess this approach will allow the "address_search" step to be made more easily generic and be put in front of any address building step.

However, I think you need to clear the enrollment.address_search after the address has been built. Otherwise, the postcode from the previous address building, will appear at the start of the next address building.

autotelik: I'm still working on the code, but yeah when to clear the search is under way and some of the edge case validations

app/factories/flood_risk_engine/steps/form_object_factory.rb
reggieb: That's not the case. It's what I'd expect from previous rails projects, but for some reason missing a translation causes an error in the app.

app/models/flood_risk_engine/address_search.rb
reggieb: Sorry - I misunderstood the meaning of the `enhancements` tag.

app/factories/flood_risk_engine/steps/form_object_factory.rb
autotelik: image

app/models/flood_risk_engine/address_search.rb
autotelik: no worries, I was pretty much making it up as I went ;-)  .. seemed like a nice colour

app/factories/flood_risk_engine/steps/form_object_factory.rb
reggieb: That's interesting! It must be a dummy app thing. This is what you get from the same page when you mount the engine via the dummy app:
missing_translation_error

reggieb: Ah! it's in the dummy apps `config/environments/development.rb`:

```
  # Raises error for missing translations
  config.action_view.raise_on_missing_translations = true
```
I'll change it to `false` as part of what I'm working on at the moment, and then drop the generic file.

autotelik: cool. thanks

autotelik:
:eyes:  ..  In meantime I'll fix travis which is VCR cassette related

app/models/flood_risk_engine/organisation.rb
```diff
@@ -12,5 +12,11 @@ class Organisation < ActiveRecord::Base
...
       unknown: 6
     }
+
+    has_one :primary_address, -> { where(address_type: Address.address_types["primary"]) },
+            class_name: "FloodRiskEngine::Address",
+            as: :addressable,
+            dependent: :restrict_with_exception
+
   end
 end
```
woodpigeon: Can an organisation have >1 address? If not we could lose the `primary_` prefix maybe (implies there may be a secondary one etc)?

app/forms/flood_risk_engine/steps/base_address_form.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      def initialize(model, enrollment)
+        @address_to_assign = Address.new
+        super(model, enrollment)
+      end
+
+    end
+  end
+end
```
woodpigeon: Like the design

app/services/flood_risk_engine/address_services/deserialize/ea_facade_to_address.rb
```diff
@@ -0,0 +1,85 @@
...
+            {
+              moniker: a["address"],
+              uprn: a["uprn"]
+            }
+          end
+
+          results
+        end
+
+        # Builds collection of attributes in correct format to simply
```
woodpigeon: its picky and I know this was copied over but could lose the `results` eg 
```
  addresses.collect do |a|
    {
      moniker: a["address"],
      uprn: a["uprn"]
    }
  end
end
```
told you it was picky ;-)

app/services/flood_risk_engine/address_services/deserialize/ea_facade_to_address.rb
```diff
@@ -0,0 +1,85 @@
...
+              blpu_state_code: a["blpu_state_code"],
+              postal_address_code: a["postal_address_code"],
+              logical_status_code: a["logical_status_code"],
+              uprn: a["uprn"]
+            }
+          end
+
+          results
+        end
+
```
woodpigeon: see above comment

app/services/flood_risk_engine/address_services/deserialize/ea_facade_to_address.rb
```diff
@@ -0,0 +1,85 @@
...
+            a.default = ""
+            {
+              easting: EaFacadeToAddress.format_xy(a["x"]),
+              northing: EaFacadeToAddress.format_xy(a["y"])
+            }
+          end
+
+          results
+        end
+
```
woodpigeon: see above comment

app/services/flood_risk_engine/address_services/search_service_failure.rb
```diff
@@ -0,0 +1,22 @@
...
+        Airbrake.notify(ex) if defined? Airbrake
+
+        Rails.logger.error "Address lookup service failed: #{ex}"
+
+        SearchServiceFailure.new(ex)
+      end
+    end
+
+  end
+end
```
woodpigeon: Nice pattern - might steal that one!

app/services/flood_risk_engine/address_services/search_service_failure.rb
```diff
@@ -0,0 +1,22 @@
+module FloodRiskEngine
+  module AddressServices
+
+    class SearchServiceFailure
+
+      attr_accessor :exception
+
+      def initialize(ex)
+        @exception = ex
+      end
```
woodpigeon: not sure `attr_accessor :exception` is used?

config/initializers/address_lookup.rb
```diff
@@ -0,0 +1,10 @@
+EA::AddressLookup.configure do |config|
+  config.address_facade_server  = Rails.application.secrets.address_facade_server
+  config.address_facade_port    = Rails.application.secrets.address_facade_port
+  config.address_facade_url     = "/address-service/v1/addresses/postcode"
+
+  # 5 was causing many timeouts in development
+  config.timeout_in_seconds     = 10
+  config.address_facade_client_id =  Rails.application.secrets.address_facade_client_id
+  config.address_facade_key       =  Rails.application.secrets.address_facade_key
+end
```
woodpigeon: yeah my 5 was a bit optimistic!

app/services/flood_risk_engine/address_services/deserialize/ea_facade_to_address.rb
autotelik: I prefer readability and protecting against future changes - where people cannot see clearly what the return value is  - over terseness

config/routes.rb
```diff
@@ -1,7 +1,9 @@
 FloodRiskEngine::Engine.routes.draw do
   resources :enrollments, only: [:new, :create] do
-    resources :steps, only: [:show, :update], controller: "enrollments/steps"
+    resources :steps, only: [:show, :update], controller: "enrollments/steps" do
+    end
```
woodpigeon: empty block not sure if intentional

autotelik: no, where i removed actions .. thanks .. fixed

.travis.yml
```diff
@@ -14,12 +14,16 @@ before_install:
...
+    - secure: atEtVos72LKeP2RkvJe3Kv775yK8YkbzdzVBv03KSWkHNqxMLFxxY03J1rJpWWyW+jEGZUjYGKtH9WIWMXqJOHkIqVkbxtvHerDBjL51sWndU6h0YQgySwXilH6nY6oWXL4ul+N/nYUiEF1sJwHDr04tUMJws7707gbZsNdN4EWELfcfpSm7GyhhPKRMNoIj1LPHrEx690EnpypGAUbM4zx55wXP8hfbXxULF+dWg43/GQVgBwi0xVxhfS17c3hMfa4Ohi8j7EacNuBcTNQLIxUcpzlNxGDpLTUfkHWNZ4soBJn6ZMyNO79Far+CsbKhqe0K08Y9wIgrz741ORixAWj7x/10Al9Q7qQ3i7R0m2ZtIbhVmFc9lYAs3U/HQI08OIOOo3uaOvwQxJDFwCqRKL1ckRfZ5Wz8ax2hT75r9w2hiJQ32RNJHswQGPEhwe0jV5Q5krumdAnI7DdRt8JbO8neugeFa4gzVPc9P1tRzZ2n755ckaGRoSu18YqxX9wUL+dIzoY9vh6r8Bkv1eaNi4kTeWl8Cy1l0ObdcpzYcsCMngDdVpmsFyr3fd3i9+dRkPQLXEaK38T6PakLA+y31I48w0lBhBCNKfi9Z1fRl92EXxlYuTqb4EdDlfxHV6GfngZVBkeNOdA5/P882SmYiQ88KF8Wlq/SD6NGW1Nuz6I=
```
woodpigeon: is it worth adding a couple of comments as to what the encrypted data is maybe in case we need to change it

woodpigeon:
Added a few small comments but looks really nice.
Happy to :ship: once @reggieb reviewed

.travis.yml
autotelik: yeah good shout .. might not even be needed if adding them in the GUI works ... I'm trying to find out why we didn't have such issues with VCR in WEX before I start blubbing like a child and asking for my mummy

app/forms/flood_risk_engine/steps/base_address_form.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      def save
+        assign_to_enrollment(address_to_assign) if address_to_assign
+
+        super
+      end
+
+      # Derived classes must implement which address instance is being managed
+      def assign_to_enrollment(_address)
+        raise NotImplementedError, "Address form cannnot set underlying Enrollment address"
```
reggieb: Typo: Three 'n' in cannot

app/services/flood_risk_engine/address_services/deserialize/ea_facade_to_address.rb
reggieb: Personally, I think @woodpigeon 's version is more readable, but that's a style thing and not something that should stop this shipping.

app/views/flood_risk_engine/enrollments/steps/_local_authority_address.html.erb
reggieb: This doesn't seem to have changed.

spec/dummy/config/initializers/address_lookup.rb
reggieb: I'm a little confused. Is this initializer over-riding the settings used in `config/initializers/address_lookup.rb`

app/models/flood_risk_engine/address_search.rb
reggieb: I don't see any clean up code. I was expecting there to be something in the LocalAuthorityAddressForm along the lines of the code below added to the end of `#assign_to_enrollment`:
```
enrollment.address_search.destroy if result
```

reggieb:
I agree with @woodpigeon this is a lot better. A couple of small things to address, and then I think it's ready to ship. Good stuff.

app/forms/flood_risk_engine/steps/base_address_form.rb
autotelik: thanks. fixed. rubymine usually good at spelling (I'm not!)

app/models/flood_risk_engine/address_search.rb
autotelik: PENDING - RIp-74 does not specify exact behaviour of back button/link
 +        # If we clear address_search now, when they go back the drop down
 +        # is cleared and shows zero results. But not sure what the designers want displayed once an address
 +        # has been saved. Maybe the dropdown should disappear and the address fields be displayed
 +        # enrollment.address_search.delete if(result)

app/models/flood_risk_engine/organisation.rb
autotelik: from the RCDP model, which inserts a Contact class between a Party/Organisation and data such as postal_address, email etc sits in Contact - as per the WEX models

app/models/flood_risk_engine/address_search.rb
reggieb: Yes, you are right. Clearing address_search will mess up the back behaviour. Perhaps we should leave this until the next address builder in the process needs to be built.

app/services/flood_risk_engine/address_services/search_service_failure.rb
autotelik: there was a previous comment about exceptions being caught and then hard to track so this is to give devs an opportunity to grab/review actual errors at external service level

app/forms/flood_risk_engine/steps/base_address_form.rb
reggieb: :smile: me neither.

app/services/flood_risk_engine/address_services/search_service_failure.rb
autotelik: oh I'm sure I already stole it :-)

app/views/flood_risk_engine/enrollments/steps/_local_authority_address.html.erb
autotelik: it didn't work all the time -  when Back or back link used for some reason it disappeared. lots of comments on apidock saying issues with prompt, so maybe that's why wex coded it this way  in first place .. hence the comment
`+<%# Create prompt manually cos this too flakey to use - e.g disappears on Back`

spec/dummy/config/initializers/address_lookup.rb
autotelik: travis and vcr are doing my nut in. experimental !

PR #69, opened 2016-05-11
Make Address and Location polymorphic

reggieb:
This will allow:

- `enrollment.location` to store grid reference and easting/northing
- `contact.address`
- `parner.address`
- `address.location`

### DIFF
--- a/app/models/flood_risk_engine/enrollment.rb
+++ b/app/models/flood_risk_engine/enrollment.rb
@@ -19,7 +19,7 @@ module FloodRiskEngine
     # The Correspondence Contact Details related to this Application a.k.a Main Contact
     belongs_to :correspondence_contact, class_name: "Contact"
 
-    has_one :location, as: :locatable
+    has_one :location, as: :locatable, dependent: :restrict_with_exception
 
     serialize :step_history, Array
 
--- a/app/models/flood_risk_engine/enrollment.rb
+++ b/app/models/flood_risk_engine/enrollment.rb
@@ -19,6 +19,8 @@ module FloodRiskEngine
     # The Correspondence Contact Details related to this Application a.k.a Main Contact
     belongs_to :correspondence_contact, class_name: "Contact"
 
+    has_one :location, as: :locatable
+
     serialize :step_history, Array
 
     before_validation :preserve_current_step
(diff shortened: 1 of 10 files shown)


app/models/flood_risk_engine/enrollment.rb
```diff
@@ -20,6 +19,8 @@ class Enrollment < ActiveRecord::Base
     # The Correspondence Contact Details related to this Application a.k.a Main Contact
     belongs_to :correspondence_contact, class_name: "Contact"
 
+    has_one :location, as: :locatable, dependent: :restrict_with_exception
```
autotelik: what is this ? the name doesn't really tell us what business need this full fills ... if this is the location where exemption activity takes place then I think `site_address` more meaningful or activity_location ...  I suspect that each Exemption maybe linked to a site so may not even be needed ... future prrofs against multiple exemptions/sites to start of use something like  EnrollmentExemption

```
class EnrollmentExemption < ActiveRecord::Base

    belongs_to :enrollment,
    belongs_to :exemption
    belongs_to :site_address,
 
```

reggieb: At the moment, the only requirement is that the location that the exemptions will refer to have a grid reference and easting/northing. All exemptions must refer to a single location (that part of the spec. see: https://eaflood.atlassian.net/browse/RIP-60). However, `exemption_location` may be a better name.

autotelik: ok. I just now we did exactly what you are doing in WEX, with site address on Enrollment and then had to change it later on 

We know bridges, and other structures likely to come along plus already have     `has_many :enrollment_exemptions` and `FloodRiskEngine::EnrollmentExemption`  .. so seems pretty easy to delegate activity_location to  the EnrollmentExemption .. but :ship:

reggieb:
I've changed `enrollment.location` as `enrollment.exemption_location`. Hopefully having locations as polymorphic it should be fairly easy to move them to the exemption (or other) object as necessary.

reggieb:
Rebased

autotelik:
Ok. shame not to just bite the bullet now, since we have the relationship to the Exemption in a separate table anyway-  but the location intrinsic to that relationship is not in that table

PR #70, opened 2016-05-11
QA issue - email validations misfiring when revisting page - RIP-119

autotelik:
_If you go back to the page and change the second email address and then continue, it doesn't validate the two fields match (although it does if you change the first email address)
Also if you change both emails and continue it shows a validation error that the emails don't match too._

Fixed by fixing method that checks if email fields have changed in the Form

### DIFF
--- a/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
+++ b/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
@@ -19,8 +19,7 @@ module FloodRiskEngine
         render :show, locals: locals
       end
 
-      # rubocop:disable Metrics/AbcSize
-      def update
+      def update # rubocop:disable Metrics/AbcSize
         success = save_form!
         if form.redirect?
           redirect_to(form.redirection_url)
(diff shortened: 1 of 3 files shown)


autotelik:
:eyes:

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -19,6 +19,7 @@ def show
         render :show, locals: locals
       end
 
+      # rubocop:disable Metrics/AbcSize
```
reggieb: I think this needs to be re-enabled after the method. Alternatively, move the error handling out into a separate method.

autotelik: agreed. disabled for single method

reggieb:
A general comment about `app/forms/flood_risk_engine/steps/correspondence_contact_email_form.rb`.

The validations are called at the class level where the `locale_key` method is available directly. So lines like this:
```ruby
validates :email_address, presence: {
          message: I18n.t("#{CorrespondenceContactEmailForm.locale_key}.errors.email_address.blank")
        }
```
can be written as:
```ruby
validates :email_address, presence: {
          message: I18n.t("#{locale_key}.errors.email_address.blank")
        }
```

reggieb:
:ship: it.

autotelik:
r.e General comment .. this is not actually the case, I've tried that but will cause a no method error as something like the ReFormGroup class is self in this location

PR #71, opened 2016-05-11
Fix problems found in testing RIP 118

reggieb:
See https://eaflood.atlassian.net/browse/RIP-118

Main issues:

- The form label was not associated with the input field
- Issues with going back in some browsers.


reggieb:
The back button behaviour occurred because the `session[:error_params]` wasn't being cleared on successful form submission.

PR #74, opened 2016-05-13
Obfuscate the enrollment id in urls

woodpigeon:
Replace the enrollment id in urls with a base58 token in order to
prevent someone hand crafting a url by guessing the id.


woodpigeon:
@autotelik @reggieb the code and logic here lifted from WEX. Only had to tweak a couple of tests. If you need to hand crank a url string, use enrollment.to_param. All other times just use just using `enrollment` in any url helper and it will call to_param implicity. 
You will need to trash your existing enrollments as only new ones will be created with a token. I haven't written a migration to do this.
The has_secure_token stuff in here is ripped from https://github.com/robertomiranda/has_secure_token which in turn cam from Rails 5. I copied it into the code base because
- a while back there was talk of dropping base58 from the has_secure_token because MySQL users might not have case sensitivity, and I didn't want a situation where we upgraded the gem and found we had less secure tokens being generated overnight!
- we generate 32 not 24 character length tokens, for extra paranoid security. There is no option to override this in the has_secure_token gem.
There are a couple of options to clean up FRE code and make the solution more usable 
- fork https://github.com/robertomiranda/has_secure_token and add a config section so we can define a custom token length - possibly submit this as a pull request on the upstream repo. Would need to rename the gem if pushing to rubygems
- just create our own gem e.g. ea-has-secure-token
- ?
OK I've clearly had too much coffee today. Thoughts welcome though.

app/models/flood_risk_engine/enrollment.rb
```diff
@@ -36,6 +39,10 @@ class Enrollment < ActiveRecord::Base
     # can be run after the state machine class has been changed
     validate :step_defined_in_state_machines
 
+    def to_param
+      token
+    end
+
     def business_type
       :foo
```
autotelik: while we are here .. what is this business_type code doing ?

autotelik:
looks good to me :ship:

reggieb:
I'm still a little surprised that none of the standard [SecureRandom](http://ruby-doc.org/stdlib-2.3.1/libdoc/securerandom/rdoc/Random/Formatter.html) methods suffice, but hey ho. 

:ship: it.

woodpigeon:
@reggieb I see your point, though this is just mimicking what we would do we if we were using [Rails 5's has_secure_token](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/secure_token.rb), so it is an emerging standard. When the project moves to Rails 5(?!), if should be possible to remove most of code in this PR.

PR #76, opened 2016-05-16
Feature/rip 57 grid reference - fix bugs found in testing

reggieb:
Some grid references that matched the test pattern were still invalid. So now using OsMapRef errors to detect if grid reference is valid.

There was a label/input field miss-match on the page.

A polyfil was needed to give the [details tag](http://www.w3schools.com/tags/tag_details.asp) cross browser compatibility.

### DIFF
--- a/app/validators/flood_risk_engine/grid_reference_validator.rb
+++ b/app/validators/flood_risk_engine/grid_reference_validator.rb
@@ -2,18 +2,51 @@
 module FloodRiskEngine
   class GridReferenceValidator < ActiveModel::EachValidator
 
-    def validate_each(record, attribute, value)
-      return true if allow_blank && value.blank?
-      OsMapRef::Location.for(value).easting
-    rescue OsMapRef::Error => e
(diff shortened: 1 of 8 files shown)


app/validators/flood_risk_engine/grid_reference_validator.rb
```diff
@@ -0,0 +1,52 @@
+# Validate that a grid reference can be processed.
+module FloodRiskEngine
+  class GridReferenceValidator < ActiveModel::EachValidator
```
autotelik: not sure if there were patterns missed in WEX but this was what we used and passed QA at the time

GRID_REFERENCE_REGEX = /\A[a-zA-Z]{2}\d{6,10}\z/i

reggieb: There is a requirement for 10 number grid references in the project. Also spaces between the groups of letters and numbers are valid, handled by OsMapRef, so I think we might as well use them.

woodpigeon:
:ship: it

PR #77, opened 2016-05-16
Feature/rip 136 manual address page

reggieb:
Provides two mechanisms by which an address can be manually modified or generated.

Use `new` to create an address (`enrollment`, `postcode` and `association` required)

Use `edit` to update an existing address (`enrollment` and `address` required)


reggieb:
Currently, the address system works going forward through the work flow. Things become "interesting" when one starts going backwards. There is an inconsistency between the back link and browser back button behaviours if one has used the manual address option. 

There are also issues as to how the change postcode button will work.

autotelik:
>>There are also issues as to how the change postcode button will work.

Do you mean once an address selected and you go back ?

reggieb:
On the manual address page, a change postcode button should exist. The problem is that there is a requirement to be able to go to the manual address page from both the post code entry page (as an option if the postcode cannot be found), and the select address page (if address is not in list). So how does the manual address page know the way to the postcode select page without it being passed as a parameter to the page? If manual address comes after address select, the address select would need to know how to pass the - how to get to postcode select - data to the manual address page.

autotelik:
Not sure I can help on the routing questions, setting up suitable redirects via the forms seemed pretty easy when I initially tried that approach for the search and select pages, you always have the enrollment and the step. think all the code disappeared now but will see if I can dig it out


The last searched post code is stored in DB so will be available to an Enrollment via the AddressSearch model .. one of the reasons for asking about audit was think of possibility of storing the history of postcodes searched too , if this brings up that requirements

reggieb:
Making the back link behave the same as the browser back button solved my problems!

app/assets/javascripts/flood_risk_engine/application.js
```diff
@@ -12,3 +12,11 @@
 //
 //= details.polyfill
 //= require_tree .
+
+
+$(function() {
```
autotelik: this is solvable with JS, not sure how this translates to your state machine but we were able to ID from the params when Back had been clicked and then work out what the available  :back transition was for the enrollment and use that transition to step back to the right state .. code for the main routine

```
    def validate_state_pre_edit
      # we bust the browser cache to ensure forms are reloaded but this causes Browser back button to behave
      # oddly or not at all
      # So we need to identify when user has gone back one state, without rendering the form
      # while trying to keep checks/code called down to a minimum

      if(params[:rendered_state].nil? && params[:state] && back_button_param_list.all? {|p| params.key?(p) })

        transitions = @enrollment.transitions_for

        back_event = transitions.find { |t| t.event == :back }

        if(back_event && back_event.from == @enrollment.state && back_event.to == params[:state])
          logger.debug("User CLICKED back Button - resetting state to [#{params[:state]}] from #{@enrollment.state}")
          @enrollment.update!(under_review: false) if(@enrollment.under_review?)
          @enrollment.update!(state: params[:state])
        end

      end
    end
```

app/controllers/flood_risk_engine/enrollments/addresses_controller.rb
```diff
@@ -0,0 +1,107 @@
+require_dependency "flood_risk_engine/application_controller"
+
+module FloodRiskEngine
+  module Enrollments
+    class AddressesController < ApplicationController
```
autotelik: Why do we need a whole Controller just for  the manual address page ? especially given the insistence that the address service lookup stuff stick to the established StepController form/show/update cycle

app/controllers/flood_risk_engine/enrollments/addresses_controller.rb
```diff
@@ -0,0 +1,107 @@
...
+        else
+          edit_enrollment_address_path(
+            enrollment,
+            address,
+            check_for_error: true
+          )
+        end
+      end
+
+      def save_form!
```
autotelik: All looks a bit similar to Step Controller code .. can this be DRY'ed into a base class or module for inclusion ?

app/helpers/flood_risk_engine/enrollments/addresses_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+          addressable_type: addressable.class.to_s,
+          postcode: postcode,
+          address_type: address_type,
+          step_back_to_postcode: step_back_to_postcode
+        )
+      end
+
+      # The change postcode page should either be the current step or the previous step.
+      # set :step_back_to_postcode to true if the postcode step will be the
+      # previous step.
```
autotelik: hard to follow really what this is for, but superficially seems like the kind of thing that the state engine should be enforcing /enabling through events and the allowable transitions between states

app/views/flood_risk_engine/enrollments/addresses/edit.html.erb
```diff
@@ -0,0 +1,29 @@
+<%= link_to(
```
autotelik: Again seems to be alot of duplication with enrollment views and between new/edit etc .. looks like alot of this should be in shared partials and then rererenced

app/assets/javascripts/flood_risk_engine/application.js
reggieb: The flood risk engine state machine does that via the go_back event - and for the journey up until now has worked OK. The problem is that the manual address page is outside of the steps system.

There could also be an inconsistency if you went back with the link and then back with the browser button, as that would take you forward! 

Making both the back link and the back button do the same thing makes these problems much easier to handle, while if JavaScript is switched off the link still works, but with a couple of peculiarities.

app/controllers/flood_risk_engine/enrollments/addresses_controller.rb
reggieb: I took the manual address functionality out of the steps because this "step" is not part of the normal workflow. It seemed to be the easiest way of handling this functionality without having to do major surgery to the steps system. It should also be relatively straight forward to re-use the manual address system for the other address pages.

app/controllers/flood_risk_engine/enrollments/addresses_controller.rb
reggieb: I think inheriting from a base class will obscure the behaviour of the controller. 

At the moment, the controller functionality is visible within a single class. Adding a base class would mean that functionality would be defined in two places.

If this control becomes difficult to read or overly complicated, then a base class may be a good way to deal with those problems. I don't believe that is the case here.

app/helpers/flood_risk_engine/enrollments/addresses_helper.rb
reggieb: I agree - if manually modifying an address was part of the normal workflow - but it is not.

So a way of stepping out of the workflow to modify an address manually, and then step back in is required. 

To ensure the manual address mechanism is reusable, you have to it all the information it needs to create a new address.

app/assets/javascripts/flood_risk_engine/application.js
autotelik: ok. Yes I've noticed odd behaviour on most pages if you start using Back and back  link

app/views/flood_risk_engine/enrollments/addresses/edit.html.erb
reggieb: I've moved the header and form into partials. Errors were already in partials. So I'm a little confused as what else you are suggesting should be moved into a partial.

app/controllers/flood_risk_engine/enrollments/addresses_controller.rb
autotelik: what do you mean by 'normal' - surely it's all one workflow ? I would have thought this is just a side branch with a couple of available back transitions.

app/views/flood_risk_engine/enrollments/addresses/_form.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    )
+  %>
+</p>
+
+<!-- Manually entered address -->
+<fieldset>
+  <legend class="visuallyhidden"><%= t(".legend") %></legend>
+
+  <div class="form-group">
+    <%= f.label :premises, class: "form-label" do %>
```
autotelik: Missing form_group_and_validation error handling ?

app/views/flood_risk_engine/enrollments/steps/_local_authority_address.html.erb
```diff
@@ -11,25 +11,35 @@
...
-    <%= f.select :uprn,  options_for_select( form.address_list.collect { |a| [ a[:moniker], a[:uprn] ] }), {},
-                 { include_blank: false,
-                   class: 'form-control form-control-select',
-                   id: 'address_match_selection'
-                 }
+    <%= f.select(
+          :uprn,
+          options_for_select( form.address_list.collect { |a| [ a[:moniker], a[:uprn] ] }),
+          {
+            include_blank: t('.num_addresses_found', num:  form.address_list.count)
```
autotelik: has this been tested in lots of front-end scenarios ? As I mentioned in the comment when I tried the `prompt` suggestion it was possible to get odd behaviour - e.g disappearing prompt - when going Back

app/controllers/flood_risk_engine/enrollments/addresses_controller.rb
reggieb: This functionality is not accessed in the normal step way - you click on a link, rather than submitting a form.

It also needs to be accessible from different places - there is a requirement for the link to be available on the post code page if the post code can't be found.

Stepping out of the steps and then stepping back in seemed like the simplest way of dealing with this complexity.

app/views/flood_risk_engine/enrollments/steps/_local_authority_address.html.erb
reggieb: Two points - you used `prompt` (on my suggestion) and the behaviour of `prompt` is precisely as you describe - it only appears if the underlying attribute is empty. It will disappear if you access the form after a value has been assigned to the attribute. If the text needs to always be there, then you need to use `include_blank` and not `prompt` (my bad suggestion to use `prompt` without also suggesting `include_blank`).

Also the include_blank needs to be in the the right hash. I did test the different scenarios.

app/controllers/flood_risk_engine/enrollments/addresses_controller.rb
autotelik: >> a base class would mean that functionality would be defined in two places.

I don't understand what you mean. Using a module you define the code once e.g for `step_forward` and then use it in 2 places.

app/views/flood_risk_engine/enrollments/addresses/_form.html.erb
reggieb: Bum - that's not the first time I've left that off. Good spot. Thank you.

app/views/flood_risk_engine/enrollments/addresses/edit.html.erb
autotelik: was just suggesting that new and edit look almost identical

app/views/flood_risk_engine/enrollments/steps/_local_authority_address.html.erb
autotelik: cool

app/controllers/flood_risk_engine/enrollments/addresses_controller.rb
reggieb: Consider this example:
```ruby
class Plus
  def modify(x, y)
    x + y
  end
end

class Minus
  include MinusFoo
end

module MinusFoo
  def modify(x, y)
    x - y
  end
end 
```
If I want to understand how `Plus` works, I only need to look at the `Plus` code. 

If I want to work out how `Minus` works, I need to look at both the `Minus` code and the `MinusFoo` code.

Abstraction is often good, and can make code more readable and easier to follow. But if done too much or too soon, it can make code harder to follow.

If the MinusFoo code needs to be re-used more than twice, the case for abstraction changes. For example if we added:
```
class TakeAway
  include MinusFoo
end

class Subtract
  include MinusFoo
end
```
The case for MinusFoo is now compelling.

At the moment, I think abstraction will make the addresses controller code harder to follow, without adding a significant benefit.

If more controllers are added, then I think we should look at abstraction.

app/views/flood_risk_engine/enrollments/addresses/edit.html.erb
reggieb: OK - I see where you are coming from. 

The problem as I see it, is that the bits that are different between the two, aren't in easy places to strip out (within the form and within grid row div)

I think it's also a level of abstraction thing again. Having partials within partials can be hard to follow.

However, the back button code is the same between both views and the step show view - that needs to be extracted. Especially as all three have the same error - they use "Back" directly rather than pulling it in from a locale file!

woodpigeon:
I have lost touch with the code around addresses so once rebased and squashed, I'm happy to let @autotelik give the go ahead on shipping.

autotelik:
well I can't say I'm happy with the design given all the comments around my stuff, adding in two little methods, whereas this adds a whole new Controller and for reasons that are not clear is all outside the state machine,  but if you have nothing to add then :ship:

reggieb:
@autotelik I think in hindsight, either one or the other of us should have done both tasks. Instead we've each come at it from a different direction and ended up with a solution that neither of use would have choosen.

PR #80, opened 2016-05-17
Allow input of grid references with lower case letters

reggieb:
There is a now a requirement for lower case grid references to be accepted as input. So `st 12345 67890` will be a valid input. I've also updated OsMapRef so that it will handle lower case input.


app/validators/flood_risk_engine/grid_reference_validator.rb
```diff
@@ -38,7 +39,7 @@ def grid_reference_pattern
     end
 
     def two_letters
-      "[A-Z]{2}"
```
autotelik: seems incredibly verbose, to create a simple regex, and then not that clear what the final regex actually is

autotelik:
:ship:

reggieb:
Thank you @autotelik

PR #81, opened 2016-05-18
Remove business_type from Enrollment

woodpigeon:
This is a relic from prototyping and can be safely removed.
Some state machine unit tests rely on it to excercise branching
however so it has been is monkey-patched back into Enrollment
in the the test.


reggieb:
Good spot

:ship: it

PR #85, opened 2016-05-19
Add declartion page functionality

reggieb:
Task [RIP-78](https://eaflood.atlassian.net/browse/RIP-78)

Note that:

1. The link to the Privacy Policy page doesn't go anywhere (should be fixed as part of [RIP-100](https://eaflood.atlassian.net/browse/RIP-100))
2. The data clip has not been built ([RIP-165](https://eaflood.atlassian.net/browse/RIP-165))

### DIFF
--- a/app/helpers/flood_risk_engine/application_helper.rb
+++ b/app/helpers/flood_risk_engine/application_helper.rb
@@ -16,7 +16,7 @@ module FloodRiskEngine
     # TODO: refactor and remove this rubocop disable
     def form_group_and_validation(form, attribute, &block)
       content = block_given? ? capture(&block) : ""
-
+      classes = ["form-group"]
       options = {
         id: error_link_id(attribute),
         role: "group",
@@ -24,15 +24,12 @@ module FloodRiskEngine
       }
 
       if form && form.errors[attribute].any?
-
-        content = content_tag(:span, form.errors[attribute].first.to_s.html_safe,
+        classes << "error"
+        content = content_tag(:span,
+                              form.errors[attribute].first.to_s.html_safe,
                               class: "error-message") + content
-
-        content_tag(:div, content, options.merge(class: "form-group error"))
-
-      else
-        content_tag(:div, content, options.merge(class: "form-group"))
       end
+      content_tag(:div, content, options.merge(class: classes.join(" ")))
     end
 
     def error_link_id(attribute)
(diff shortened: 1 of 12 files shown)


app/helpers/flood_risk_engine/application_helper.rb
```diff
@@ -55,9 +55,5 @@ def page_title(title)
 
       title
     end
-
-    def submit_button_text(_enrollment)
```
autotelik: why remove this ? You may find you need it once the check details page comes along or definitly for the declaration page, as these pages require different wording of the burtton

app/views/flood_risk_engine/enrollments/steps/show.html.erb
```diff
@@ -35,7 +35,15 @@ see the DigitalServicesCore::BackLink class -->
       <%= render step.to_s, form: form, step: step, f: f %>
 
       <div class="form-group">
-        <%= f.submit submit_button_text(form.enrollment), class: "button"  %>
+        <%=
+          f.submit(
```
autotelik: this form is so much harder to read

spec/dummy/config/initializers/address_lookup.rb
```diff
@@ -8,3 +8,5 @@
   config.address_facade_client_id =  Rails.application.secrets.address_facade_client_id
   config.address_facade_key       =  Rails.application.secrets.address_facade_key
 end
+
+EA::AddressLookup.logger.level = Logger::ERROR
```
autotelik: :+1:

autotelik:
:ship:

app/helpers/flood_risk_engine/application_helper.rb
reggieb: I removed it because I think just having `t("global.continue")` in the places where it is needed is clearer and more concise. The method is only used in two places. Also, have a look at app/views/flood_risk_engine/enrollments/steps/show.html.erb within this PR.

app/views/flood_risk_engine/enrollments/steps/show.html.erb
reggieb: Is this better? - I've added a `step_t` method that I think tidies this up.

autotelik: personally wouldn't change long established code, for no obvious benefit especially with the lack of decent feature and view specs but :ship:

PR #87, opened 2016-05-20
Implement Address selection from dropdown - RIP-135

autotelik:
Bug fixes around address validation

### DIFF
--- a/app/forms/flood_risk_engine/steps/base_address_form.rb
+++ b/app/forms/flood_risk_engine/steps/base_address_form.rb
@@ -49,11 +49,16 @@ module FloodRiskEngine
         if address_attributes
           validator = AddressValidator.new(address_attributes)
 
-          @address_to_assign = Address.new(address_attributes) if validator.validate
+          if validator.validate
+            @address_to_assign = Address.new(address_attributes)
+            logger.debug("LocalAuthority address now #{@address_to_assign.inspect}")
+          else
+            # not too sure how to merge this validation into Form automatically so for now
+            # manually clone error messages
+            logger.debug("Address Valid #{validator.valid?} Errors [#{validator.errors.inspect}]")
 
-          logger.debug("AddressValidator returned [#{validator.valid?}]")
-
-          logger.debug("LocalAuthority address now #{@address_to_assign.inspect}")
+            validator.errors.each { |_e, m| errors.add :address, m }
+          end
 
           validator.valid?
         else
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,6 @@ group :test do
   gem "rspec-html-matchers"
   gem "database_cleaner", "~> 1.5"
   gem "codeclimate-test-reporter", require: false
-  gem "vcr", "~> 3.0"
+  gem "vcr", "~> 3.0.1"
   gem "webmock", "~> 1.24"
 end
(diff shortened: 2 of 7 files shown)


autotelik:
:eyes:

woodpigeon:
When the address service is misconfigured eg `ADDRESS_FACADE_TEST_SERVER = "yay.com"`
then there is a `NoMethodError (undefined method `empty?' for nil:NilClass):` error at `ea_facade_to_address.rb:10`
Can be fixed by replacing 
```return true if inbound && inbound.is_a?(Hash) && !inbound["results"].empty?```
with
```return true if inbound && inbound.is_a?(Hash) && !inbound["results"].blank?```

**Comment by autotelik** [comment]
Ok I'll have a look .. cos seems odd we get to that point if its misconfigured .. seems we should have a SearchServiceFailure before we get to  that point

**Comment by woodpigeon** [comment]
To get specs to pass I had to make 2 changes:

1.  Raise VCR errors in test so I know why the test is failing e.g. in `SearchServiceFailure`
```
def self.build(ex)
  trap_vcr_errors_due_to_missing_env_vars(ex) if Rails.env.test?

  Airbrake.notify(ex) if defined? Airbrake

  Rails.logger.error "Address lookup service failed: #{ex}"
  Rails.logger.debug("EA::AddressLookup Config was: #{EA::AddressLookup.config.inspect}") if Rails.env.test?

  SearchServiceFailure.new(ex)
end

def self.trap_vcr_errors_due_to_missing_env_vars(ex)
  raise ex if ex.message =~ /VCR/
end
```

2. Stub out #search in `postcodes_spec.rb:117` to stop a missing VCR cassette error (not sure why this works for anyone else ;-) )

```
expect_any_instance_of(AddressServices::FindByPostcode).to receive(:search).and_return(nil)
expect_any_instance_of(AddressServices::FindByPostcode).to receive("success?").and_return(false)
```

**Comment by autotelik** [comment]
VCR errors are already fully recorded in test.log .. I usually look there when a test fails and not clear why

File: app/forms/flood_risk_engine/steps/base_address_form.rb [hunk 1]
```diff
@@ -49,11 +49,16 @@ def assign_via_uprn
...
-          logger.debug("LocalAuthority address now #{@address_to_assign.inspect}")
+          if validator.validate
+            @address_to_assign = Address.new(address_attributes)
+            logger.debug("LocalAuthority address now #{@address_to_assign.inspect}")
+          else
+            # not too sure how to merge this validation into Form automatically so for now
+            # manually clone error messages
+            logger.debug("Address Valid #{validator.valid?} Errors [#{validator.errors.inspect}]")
+
+            validator.errors.each { |_e, m| errors.add :address, m }
```
reggieb: If errors returns a hash, iterating over the values would remove the need to have the unused `_e` variable:

`validator.errors.values.each { |message| errors.add :address, message }`

woodpigeon:
@autotelik I have added those changes around vcr error in another branch so could be merged in from there, so don't worry. The VCR errors I had were due to my misconfiguration of the address facade, but then `postcodes_spec.rb:117` was complaining about a missing cassette as it does go off-box. Anyway ignore as solved elsewhere and its not stopping CI tests from passing.

woodpigeon:
:ship: it when happy

PR #91, opened 2016-05-24
Confirmation page - RIP-80 

autotelik:
Add a ref number field to enrollment, migration and after_create hook to populate it
Make display of continue button configurable
Create bare bones Service object to handle finalize once journey  complete


app/forms/flood_risk_engine/steps/confirmation_form.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+
+      private
+
+      def initialize(enrollment)
+        logger.info("#{enrollment.inspect} #{enrollment.correspondence_contact.inspect}")
+
+        super enrollment
+
+        @no_header_in_show = true
```
reggieb: I'd suggest defining the no_header_in_show method in this form so that it always returns true:

```
def no_header_in_show
  true
end
```

app/views/flood_risk_engine/enrollments/steps/show.html.erb
```diff
@@ -3,10 +3,21 @@
...
+      <%= render(
+              "flood_risk_engine/enrollments/header",
+              title_text: t("flood_risk_engine.enrollments.steps.#{step}.heading")
+          )
+      %>
+    <% end %>
+
+    <%= render 'flood_risk_engine/enrollments/validation_errors', form: form %>
+
+    <% # Using 'as:' changes the name of the fields in the form from eg
```
reggieb: This puts the comment code about `as:` back in after it was replaced with a shorter comment on the line using the `:as` (line 25). If you prefer the commend at it was, can you please remove the comment on line 25.

autotelik: thanks, there was bit of a funny merge on this, so I'll revisit

app/views/flood_risk_engine/enrollments/steps/_confirmation.html.erb
```diff
@@ -1 +1,35 @@
...
+<p class="text"><%= t ".email_sent", email: form.email %></p>
+
+<h2 class="heading-medium">
+  <%= t('.heading_h2') %>
+</h2>
+
+<p class="text"><%= t('.para_text2') %></p>
+<p class="text"><%= t('.para_text3') %></p>
+<p class="text"><%= t('.para_text4') %></p>
+<p class="text"><%= t('.para_text5') %>
```
reggieb: There is a problem with the pattern: `translation_one`, `link`, `translation_two` in that it is difficult to manage across different languages. 

For example if you wanted a link on "you" in "Who are you?", in English the link would go at the start of the string. But if you look at the equivalent Italian, "Tu chi sei", the link would have to be around "Tu" at the start of the string.

Therefore, I think the pattern where you pass the link into the translation as a variable, is a much better pattern. This will allow you to clean up the locale file (see above) and the code here:

```
<p class="text">
<%= t('.para_text5',
    link: link_to('flood risk activities guidance', t('.activities_url'), target: "_blank")
    ).html_safe
%>
</p>
```

app/views/flood_risk_engine/enrollments/steps/show.html.erb
autotelik: Ok yeah I cocked up the merge, that  longer comment should have been removed and the duplicated validation render .. fixed, TVM!

config/locales/flood_risk_engine/enrollments/steps/_confirmation.en.yml
```diff
@@ -0,0 +1,30 @@
...
+          heading_h2: What happens next
+          heading_h2_1: If your activity changes
+          para_text: 'Your reference is:'
+          email_sent: We've sent a confirmation email to %{email}
+          para_text2: You can begin work on the activity covered by your exemptions.
+          para_text3: Exemptions don’t expire, except for those for dredging which have time limits.
+                      If want to do the same activity again you must register a new exemption.
+          para_text4: We may check to see if your activity matches the conditions and
+            description of your exemption.
+          para_text5: You can check the conditions and description of your exemption
```
reggieb: I think:

para_text5: "You can check the conditions and description of your exemption in the %{link}."

is a better pattern

spec/forms/flood_risk_engine/confirmation_form_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      let(:form) { ConfirmationForm.factory(enrollment) }
+
+      subject { form }
+
+      it_behaves_like "a form object"
+
+      it { is_expected.to be_a(ConfirmationForm) }
+
+      describe "Save" do
+        include FloodRiskEngine::Engine.routes.url_helpers
```
reggieb: Do you need to include `FloodRiskEngine::Engine.routes.url_helpers` here?

app/views/flood_risk_engine/enrollments/steps/_confirmation.html.erb
autotelik: as an aside we don't need to support multiple languages, GDS policy

reggieb:
A couple of minor issues, otherwise :ship: it.

app/views/flood_risk_engine/enrollments/steps/_confirmation.html.erb
reggieb: Sorry - working with multiple languages is where this issue has caught me out in the past, and is the simplest example to show what the problem is. I realise that a second language is unlikely to be needed, but can't help seeing this as something that can be done better.

Feel free to ignore.

autotelik: Fixed

autotelik: No I like that pattern, changes made, the aside was just an aside :-)

PR #99, opened 2016-05-25
Add FRA23 and FRA24 to exemptions.

reggieb:
See [RIP-158](https://eaflood.atlassian.net/browse/RIP-158)

Note: that this requires a `rake db:seed` before the change will take effect. Seeding can be run multiple times without creating duplicates.


woodpigeon:
Looks good :ship: it

PR #100, opened 2016-05-25
Remove VCR and instead mock calls to EA::AddressLookup

reggieb:
This update adds methods available in RSpec tests, to allow EA::AddressLookup methods to be mocked and return data defined in spec/support/flood_risk_engine/mock_data/address_lookup.yml


woodpigeon:
Seems like a nice pattern, adding mock data to the gem.
:ship: it

PR #105, opened 2016-05-28
Remove VCR and instead mock calls to EA::AddressLookup 2

woodpigeon:
Where specs use VCR to mock EA::AddressLookup.find_* methods, replace
with the mock methods provided by the EA::AddressLookup.


app/forms/flood_risk_engine/steps/local_authority_address_form.rb
```diff
@@ -4,7 +4,9 @@ module Steps
     class LocalAuthorityAddressForm < BaseAddressForm
 
       def self.factory(enrollment)
-        raise(FormObjectError, "No Organisation set for step #{enrollment.current_step}") unless enrollment.organisation
+        unless enrollment.organisation
```
autotelik: most ruby style guides seem to prefer the former

spec/forms/flood_risk_engine/steps/local_authority_address_form_spec.rb
```diff
@@ -58,35 +40,33 @@ def form_requires_address_lookup
         }
 
         it "is valid when valid UK UPRN supplied via drop down but no street_address present" do
-          form_requires_address_no_street_lookup do
-            expect(form.validate(valid_but_no_street_address)).to eq true
-          end
+          mock_ea_address_lookup_find_by_uprn
+          expect(form.validate(valid_but_no_street_address)).to eq true
```
autotelik: I need to dig deeper to be 100% but I'm concerned that here we are no longer exercising the correct underlying test. AFAIK the data returned by the mock is always the same postcode/data set (which will validate) but this test was to check the situation where post code valid e.g BS9 9XX  but  the subsequently returned address results set is empty

spec/forms/flood_risk_engine/steps/local_authority_address_form_spec.rb
```diff
@@ -58,35 +40,33 @@ def form_requires_address_lookup
...
 
-            address = subject.enrollment.reload.organisation.primary_address
+            address = form.enrollment.reload.organisation.primary_address
 
             expected_attributes = {
               address_type: "primary",
               premises: "HORIZON HOUSE",
               street_address: "DEANERY ROAD",
-              locality: nil,
+              locality: "SOMEWHERE",
```
autotelik: seems better for test data to behave exactly as returned by address service - looks like the mock data set has been manually altered

app/forms/flood_risk_engine/steps/local_authority_address_form.rb
woodpigeon: normally agree, but my editor flagged this line as over 120 chars long so changed it

spec/forms/flood_risk_engine/steps/local_authority_address_form_spec.rb
autotelik: I think we need a smarter mock for this as the required params are 

let(:valid_but_no_street_address) {
          {
            "#{form.params_key}": { post_code: "HX3 0TD", uprn: "10010175140" }
          }
        }

autotelik: perhaps this style  more appropriate

 allow_any_instance_of(AddressServices::FindByPostcode).to receive("search").and_return([])

spec/forms/flood_risk_engine/steps/local_authority_address_form_spec.rb
woodpigeon: Good point - have change ea-address_lookup to return "" for locality as per address service result

PR #111, opened 2016-05-31
Limited company registration number page - RIP-81

autotelik:
Create generic companies house validator
Implement view, form and locale data
Rename company number, registration number, so more easily shared across organisation types

**N.B**
Stub forms/views for the two new states (following this state) so the journey in browser can still continue - therefore code for limited_company name and address should not be treated as QA ready

### DIFF
--- a/app/presenters/flood_risk_engine/enrollment_presenter.rb
+++ b/app/presenters/flood_risk_engine/enrollment_presenter.rb
@@ -38,6 +38,11 @@ module FloodRiskEngine
       title
     end
 
+    def organisation_registration_number
+      return unless enrollment.organisation && enrollment.organisation.registration_number.present?
+      enrollment.organisation.registration_number
+    end
+
     private
 
     attr_reader :enrollment
--- a/app/presenters/flood_risk_engine/tabular_enrollment_detail/row_builder.rb
+++ b/app/presenters/flood_risk_engine/tabular_enrollment_detail/row_builder.rb
@@ -65,6 +65,13 @@ module FloodRiskEngine
                   value: enrollment_presenter.correspondence_contact_telephone_number
       end
 
+      def organisation_registration_number_row
+        return unless enrollment_presenter.organisation_registration_number
+        build_row name: :company_number,
+                  value:  enrollment_presenter.organisation_registration_number,
+                  step: "#{organisation_type}_number"
+      end
+
       private
 
       # ==== Arguments
--- a/app/presenters/flood_risk_engine/tabular_enrollment_detail_presenter.rb
+++ b/app/presenters/flood_risk_engine/tabular_enrollment_detail_presenter.rb
@@ -27,10 +27,11 @@ module FloodRiskEngine
         row_builder.organisation_type_row,
         row_builder.organisation_name_row,
         row_builder.organisation_address_row,
+        row_builder.organisation_registration_number_row,
         row_builder.correspondence_contact_name_row,
         row_builder.correspondence_contact_telephone_row,
         row_builder.correspondence_contact_email_row
-      ].flatten!
+      ].flatten.compact
     end
 
     def row_builder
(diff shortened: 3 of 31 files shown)


app/presenters/flood_risk_engine/enrollment_presenter.rb
```diff
@@ -38,6 +38,11 @@ def correspondence_contact_name
       title
     end
 
+    def organisation_registration_number
+      return unless enrollment.organisation && enrollment.organisation.registration_number.present?
+      enrollment.organisation.registration_number
+    end
+
     private
```
woodpigeon: I think you can just add to the the delagate call at the top of the class - will handle nil, e.g. 
```
delegate :name,
             :primary_address,      
             :registration_number,
             to: :organisation, allow_nil: true, prefix: true
```

app/presenters/flood_risk_engine/tabular_enrollment_detail/row_builder.rb
```diff
@@ -65,6 +65,13 @@ def correspondence_contact_telephone_row
...
       end
 
+      def registration_number_row
+        return unless enrollment_presenter.organisation_registration_number
+        build_row name: :company_number,
+                  value:  enrollment_presenter.organisation_registration_number,
+                  step: "#{organisation_type}_number"
+      end
+
       private
```
woodpigeon: I wonder if organisation_registration_number is more consistent? Otherwise its a bit ambiguous as to what a registration is.

app/presenters/flood_risk_engine/tabular_enrollment_detail_presenter.rb
```diff
@@ -27,10 +27,11 @@ def build_rows
         row_builder.organisation_type_row,
         row_builder.organisation_name_row,
         row_builder.organisation_address_row,
+        row_builder.registration_number_row,
         row_builder.correspondence_contact_name_row,
```
woodpigeon: `organisation_registration_number_row` feels more consistent and unambiguous here as the other org rows have that prefix

app/validators/ea/validators/companies_house_number_validator.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      def validate_each(record, attribute, value)
+        value.strip!
+        unless value =~ VALID_COMPANIES_HOUSE_REGISTRATION_NUMBER_REGEX
+          record.errors.add attribute, (options[:message] || :invalid)
+        end
+      end
+    end
+
+  end
```
woodpigeon: 👍

config/initializers/address_lookup.rb
```diff
@@ -4,7 +4,7 @@
   config.address_facade_url     = "/address-service/v1/addresses/postcode"
 
   # 5 was causing many timeouts in development
-  config.timeout_in_seconds     = 10
-  config.address_facade_client_id =  Rails.application.secrets.address_facade_client_id
-  config.address_facade_key       =  Rails.application.secrets.address_facade_key
+  config.timeout_in_seconds       = Rails.application.secrets.address_facade_timeout
+  config.address_facade_client_id = Rails.application.secrets.address_facade_client_id
+  config.address_facade_key       = Rails.application.secrets.address_facade_key
```
woodpigeon: no suggestions, but have realised the engine is making assumptions here that the host app will have a secrets file with those settings - not a big issue for us as we control the host apps - will just need to make sure fo and bo and dummy all have that setting.

woodpigeon:
See what you think re `organisation_registration_number_row` and the `delegate` change above, then :ship: it

app/presenters/flood_risk_engine/enrollment_presenter.rb
autotelik: I do like delegation in places but the above feels much harder to read and understand exactly what is going on than just a plain method

config/initializers/address_lookup.rb
autotelik: yeah I agree, it's slightly odd pattern to me, IMHO not sure what secrets is giving us here, why we don't just use the ENV directly in the initializer, given that all the secrets does is assign ENV['ADDRESS_FACADE_TIMEOUT']

PR #113, opened 2016-06-01
Make dredging length input numeric

woodpigeon:
Leave the database column as a string in case we wish to revert to free-text.
Length should be between 1 and 1500 but configurable.

### DIFF
--- a/spec/forms/flood_risk_engine/steps/grid_reference_form_spec.rb
+++ b/spec/forms/flood_risk_engine/steps/grid_reference_form_spec.rb
@@ -29,6 +29,12 @@ module FloodRiskEngine
         }
       end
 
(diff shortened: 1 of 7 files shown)


spec/forms/flood_risk_engine/steps/grid_reference_form_spec.rb
```diff
@@ -155,15 +161,26 @@ module Steps
           end
 
           context "with too long a dredging_length" do
-            let(:dredging_length) { Faker::Lorem.characters(26) }
+            let(:dredging_length) { "150,000" }
```
autotelik: comma looks odd

autotelik:
:ship:

PR #120, opened 2016-06-03
Fix bug in regexp logic for company name

autotelik:
Simplify regexp matching
Resolves bug where a single special char raised invalid (even though that's a valid name)

### DIFF
--- a/app/validators/ea/validators/companies_house_name_validator.rb
+++ b/app/validators/ea/validators/companies_house_name_validator.rb
@@ -11,8 +11,10 @@ module EA
 
     class CompaniesHouseNameValidator < ActiveModel::EachValidator
 
+      VALID_COMPANIES_HOUSE_NAME_REGEX = Regexp.new(/[\^|_~¬`]/).freeze
+
       def validate_each(record, attribute, value)
-        if value !~ CompaniesHouseNameValidator.valid_name_regex
+        if VALID_COMPANIES_HOUSE_NAME_REGEX.match(value)
           record.errors.add(
             attribute,
             (options[:message] || I18n.t("ea.validation_errors.companies_house_name.#{attribute}.invalid"))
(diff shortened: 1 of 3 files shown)


autotelik:
:eyes:

app/validators/ea/validators/companies_house_name_validator.rb
```diff
@@ -11,8 +11,10 @@ module Validators
 
     class CompaniesHouseNameValidator < ActiveModel::EachValidator
 
+      VALID_COMPANIES_HOUSE_NAME_REGEX = Regexp.new(/[\^|_~¬`]/).freeze
```
reggieb: I don't think you need the `Regexp.new`. 

```
/[\^|_~¬`]/.freeze 
```

should work.

app/validators/ea/validators/companies_house_name_validator.rb
```diff
@@ -24,8 +26,8 @@ def self.name_max_length
         160
       end
 
-      def self.valid_name_regex
-        /\A[\p{Alpha}\p{N}][^\^|_~¬`]+\z/i
+      def self.disallowed_chars
+        @disallowed_chars ||= ["^", "|", "_", "~", "¬", "`"]
```
reggieb: I'd suggest this is a little clearer: 
```
@disallowed_chars ||= %w( ^ | _ ~ ¬ ` )
```

spec/validators/companies_house_name_validator_spec.rb
```diff
@@ -14,33 +14,39 @@ class NameValidatable
...
+      end
+
+      context "With only a single special char" do
+        let(:name) { "@" }
+        it "is valid" do end
+      end
+
+      context "With only multiple special chara" do
+        let(:name) { "@ $$" }
+        it "is valid" do end
```
reggieb: The tests with `it "is valid" do end` aren't testing anything. If they are not needed, they should be removed.

reggieb: I think the use of an after block is confusing, as it forces the need for the strange `it "is valid" do end` entries. Would `is_expected.to be_valid` in each context be simpler?

autotelik:
At this stage I'm not changing this code as its intentions are clear to me

PR #131, opened 2016-06-08
Individual address - RIP-88

autotelik:
https://eaflood.atlassian.net/browse/RIP-88

Implement address pages for Individual	

**N.B** This PR lays the basis for Individual address but does not complete story RIP-88  - pushing now so the Individual journey can be completed - which enables RIP-87 Individual Name to go into QA

### DIFF
--- a/app/views/flood_risk_engine/enrollments/steps/_add_exemptions.html.erb
+++ b/app/views/flood_risk_engine/enrollments/steps/_add_exemptions.html.erb
@@ -1,6 +1,7 @@
 <p class="panel panel-border-wide">
   <%= t("flood_risk_engine.enrollments.steps.add_exemptions.lead_paragraph.text_1") %>
-  <br></br>
+  <br>
+  </br>
   <%= t("flood_risk_engine.enrollments.steps.add_exemptions.lead_paragraph.text_2") %>
 </p>
 
@@ -11,7 +12,7 @@
   <%= form_group_and_validation(form, :exemption_ids) do %>
     <%= f.collection_radio_buttons(:exemption_ids, form.all_exemptions, :id, :code) do |b|
       b.label(class: "block-label exemption") do
-        b.radio_button(data: {code: b.object.code}) +
+        b.radio_button(data: {code: b.object.code}, 'aria-expanded': "false") +
             b.object.summary + ' ' + content_tag('span', b.text, class: "exmpt")
       end
     end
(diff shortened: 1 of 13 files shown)


app/forms/flood_risk_engine/steps/local_authority_address_form.rb
```diff
@@ -15,28 +15,10 @@ def self.params_key
...
       def initialize(model, enrollment)
         super(model, enrollment)
-
-        # Consensus is that the original drop down list should be displayed when a user clicks Back,
-        # If address fields required local_authority_address_back can include partials 'shared/address_fields'
-        # or 'shared/display_address'
-        partial = "flood_risk_engine/enrollments/steps/local_authority_address_back"
-
-        @partial_to_render = partial if enrollment.organisation.primary_address.present?
       end
```
woodpigeon: I think this is the only place @partial_to_render is set  - does this mean its going? If so should get rid of other `partial_to_render` references eg in  steps/base_form.rb and enrollments/steps/show.html.erb

app/views/flood_risk_engine/enrollments/steps/_individual_address.html.erb
```diff
@@ -0,0 +1 @@
+<%= render 'flood_risk_engine/shared/address', form: form, step: step, f: f %>
```
woodpigeon: I am guessing other *_address.html.erb steps should use this DRY approach now? If so might be worth creating a github issue to make sure those changes are picked up?

app/forms/flood_risk_engine/steps/local_authority_address_form.rb
autotelik: Yes agreed, I don't think requirements will change now so no need to potentially render anything different to original view

To complete I will definitely rationalise address forms and partials to use new base classes

app/views/flood_risk_engine/enrollments/steps/_individual_address.html.erb
autotelik: Done : https://github.com/EnvironmentAgency/flood-risk-engine/issues/134

app/views/flood_risk_engine/enrollments/steps/_add_exemptions.html.erb
```diff
@@ -1,6 +1,7 @@
 <p class="panel panel-border-wide">
   <%= t("flood_risk_engine.enrollments.steps.add_exemptions.lead_paragraph.text_1") %>
-  <br></br>
+  <br>
+  </br>
```
reggieb: [`<br>` is a self closing tag.](http://stackoverflow.com/questions/4288460/html-whats-the-correct-form-of-br) So it should be either `<br>` or `<br><br>' (which is two breakes).

PR #137, opened 2016-06-09
Address  + Postcode - RIP-88

autotelik:
Rationalise duplicated partials for all OrgType's with an address/postcode partial


autotelik:
:eyes:

app/views/flood_risk_engine/shared/_address.html.erb
```diff
@@ -1,14 +1,14 @@
-<%= t('.postcode_hint') %>
+<%= form.t('.postcode_hint') %>
 <p>
```
woodpigeon: Is the `form.t()`  so translations are scoped to the step's yml file and not a generic address.en.yml?

woodpigeon:
Looks good to me but @reggieb might want to review because of potential impact on his address steps eg other?

autotelik:
I  have used  @reggieb version of the partials so should be no regression - only rationalised the location

app/views/flood_risk_engine/shared/_address.html.erb
autotelik: not entirely sure TBH, this is from the @reggieb version, and as it all worked as is  - without any locale changes to my stuff  - I assume it degrades  ok to generic stuff when no step yml defined

woodpigeon:
@autotelik ok to :ship: then

PR #141, opened 2016-06-10
Feature/rip 122 prevent url sharing

reggieb:
[RIP-122](https://eaflood.atlassian.net/browse/RIP-122)

I think this may need some more thought. I've implemented what I think is a simple solution, but:

1. we will need to over-ride the behaviour in the back office app, if NCCC staff are going to continue users' journeys.
2. It is based on session, so won't allow someone to close their browser and reopen it to return to the journey.
3. Will stop users returning to the completed step after they've closed the browser.
4. Won't allow them to work on two journeys at the same time.

I'm tempted to store the journey in an array - as that will handle case 4, and change the store to a cookie as that will handle case 2 and 3 (but will lose the benefit of having the key in the database session store rather than the browser). Alternatively 3 could be addressed by ignoring the check once the journey is completed.

Can someone please have a look through as I could do with a second pair of eyes on this.


woodpigeon:
Code looks good but you are right to raise the caveats.
1) should be easy to override in a BO implementation of `check_journey_valid` I guess so all good
2) Not sure that's a bit problem but we might need to run it past the business - though your cookie idea might be better
3) As 2, but I think that's ok - the process is complete, and they will have an email. But yes using a cookie might be better
4) One for the business to think about maybe, but your array idea is not a good one.
Sorry - not very constructive comments! 
I would be tempted to explore the dedicated cookie approach and see if it feels better that way. A the end of the day we are trying (actually a bit reluctantly) to stop sharing the link across browsers, so being too restrictive might backfire.

reggieb:
I guess I'll need to encrypt or otherwise hash the cookie as if this is being done for security reasons, there's no point in using an easily read and understood cookie value.

OK - I'll try that.

woodpigeon:
Good point. We have a story to make all cookies secure so are only transmitted over https, but that's not solving the problem thinking about it, so yeah a signed cookie maybe using secret_key_base is probably a good idea

reggieb:
I've moved the journey token to a cookie (that expires in a day, but should be easy to change that - to use an environment variable?), and applied simple encoding to it so it's not obvious how to spoof it.

woodpigeon:
Great stuff

woodpigeon:
Re expiration, I like your idea of tyring an env variable first, maybe defaulting to 1 day if not found.

app/services/flood_risk_engine/simple_encoding.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def characters
+      "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
+    end
+
+    def encoding
+      "N8nGfS1rpcHvmoiP6WB9FMQwXjyZ0Rhu73stKAla2Yd4LTJqkeUzgEOxbCD5VI"
+    end
+  end
+end
```
woodpigeon: Would be be simpler to use SECRET_KEY_BASE here? Its always available and is what's used to encode the session cookie.

woodpigeon: Youve probably looked at this but you can use a signed cookie which will automatically sign it using secret key base - though actually I'm not sure if its actually encrypted too, haven't checked
from http://api.rubyonrails.org/classes/ActionDispatch/Cookies.html
```
# Sets a signed cookie, which prevents users from tampering with its value.
# The cookie is signed by your app's `secrets.secret_key_base` value.
# It can be read using the signed method `cookies.signed[:name]`
cookies.signed[:user_id] = current_user.id
```

reggieb: Yes - that looks like a better solution. I looked at setting cookies to be encrypted by default, but the setting I was using interferred with the operation of the database session store, so went for the basic encoding option. Oh and can't use SECRET_KEY_BASE as just a cipher - so needs to include all the characters.

woodpigeon:
:ship: ahoy

PR #143, opened 2016-06-13
Limited liability partnership name page - RIP-85 

autotelik:
https://eaflood.atlassian.net/browse/RIP-85

Seems like most of work done on basic LLP journey as part of 84 is fine so this is just related rspecs
for 85


woodpigeon:
:ship: it from me although the commit message is far too verbose :smiley:

PR #145, opened 2016-06-13
Add partnership step - where the name of a partner is entered.

reggieb:
See [RIP-89](https://eaflood.atlassian.net/browse/RIP-89)

I'll need to return to this page to ensure that when a user is sent back to this page, it allows them to add an additional partner (as in WEX). I have added some functionality in preparation for this (`form.partner_count`)

### DIFF
--- a/app/forms/flood_risk_engine/steps/partnership_form.rb
+++ b/app/forms/flood_risk_engine/steps/partnership_form.rb
@@ -1,0 +1,48 @@
+module FloodRiskEngine
+  module Steps
+    class PartnershipForm < BaseForm
+      def self.factory(enrollment)
+        contact = Contact.new
+        new(contact, enrollment)
+      end
+
+      def self.params_key
+        :partnership
+      end
+
+      def self.max_length
+        70
+      end
+
+      # Used to work out whether this is the first partner being created.
+      # The count is effectively, the number of partners already built,
+      # plus the one being built. So the total number of partners.
+      # Counting this way makes the translations easier (:one and :other syntax)
+      def partner_count
+        enrollment.organisation.partners.count + 1
+      end
+
+      property :full_name
+
+      validates(
+        :full_name,
+        presence: {
+          message: t(".errors.full_name.blank")
+        },
+        "flood_risk_engine/text_field_content" => {
+          allow_blank: true
+        },
+        length: {
+          maximum: max_length,
+          message: t(".errors.full_name.too_long", max: max_length)
+        }
+      )
+
+      def save
+        super
+        enrollment.organisation.partners.create(contact: model)
+      end
+
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


app/forms/flood_risk_engine/steps/partnership_form.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      def self.max_length
+        70
+      end
+
+      # Used to work out whether this is the first partner being created.
+      # Needs to start at one for translations (:one and :other syntax)
+      def partner_count
+        enrollment.organisation.partners.count + 1
```
autotelik: this seems a bit odd .. when you have no partners, your partner count is 1 ?  maybe  method badly named

app/models/flood_risk_engine/organisation.rb
```diff
@@ -2,6 +2,7 @@ module FloodRiskEngine
   class Organisation < ActiveRecord::Base
     belongs_to :contact
     has_one :enrollment, dependent: :restrict_with_exception
+    has_many :partners
```
autotelik: since most Organisations have nothing to do with partners this smells

app/forms/flood_risk_engine/steps/partnership_form.rb
reggieb: The count is effectively, the number of partners persisted in the database, plus the 1 you are currently building. I'll update the comment to that effect.

app/models/flood_risk_engine/organisation.rb
reggieb: I can't see an alternative that adds any value without adding complication and/or abstraction. I'm open to suggestions.

autotelik:
:ship:

app/models/flood_risk_engine/organisation.rb
woodpigeon: I agree with @autotelik this is a bit misleading. In WEX we subclassed all org types using STI but in flood this is the only piece of code that is org-class specific like this, so that approach a bit overkill. I can't think of a solution but we could add a comment to make it clear it applies only when the Organisation is a partnership. There are conditions on a has_many (see second answer [here](http://stackoverflow.com/questions/9556153/is-it-possible-to-create-a-conditional-association-in-model)) but that's more about scoping I suppose, and isn't going to make that relationship disappear for non partnership Organisations.

woodpigeon:
:ship: it when happy

app/models/flood_risk_engine/organisation.rb
autotelik: Too late and irrelevant now in big scheme of things .. but not sure why STI was over kill, when its the correct abstraction ?  Coming from a C++/OO background it seemed very straight fwd to implement and understand. The enum adds complexity here with no benefit in managing behaviour.

reggieb: Use a `where` clause on the `has_many` call doesn't work, because the `where` is applied to the class being joined to, not the model hosting the `has_many` declaration.

STI would make sense if the variation between the behaviours of different organisation types was greater than this one instance. Adding 6 class object to handle one small variant is overkill.

PR #152, opened 2016-06-14
Feature/rip 92 add partner address pages

reggieb:
Includes postcode and address pages. [RIP-92](https://eaflood.atlassian.net/browse/RIP-92)

### DIFF
--- a/app/models/flood_risk_engine/enrollment.rb
+++ b/app/models/flood_risk_engine/enrollment.rb
@@ -11,7 +11,7 @@ module FloodRiskEngine
     # we query never from contact to its enrollment
     belongs_to :applicant_contact, class_name: "Contact"
     belongs_to :organisation
-    delegate :org_type, to: :organisation, allow_nil: true
+    delegate :org_type, :partners, to: :organisation, allow_nil: true
 
     has_many :enrollment_exemptions,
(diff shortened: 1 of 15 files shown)


app/models/flood_risk_engine/enrollment.rb
```diff
@@ -11,7 +11,7 @@ class Enrollment < ActiveRecord::Base
     # we query never from contact to its enrollment
     belongs_to :applicant_contact, class_name: "Contact"
     belongs_to :organisation
-    delegate :org_type, to: :organisation, allow_nil: true
+    delegate :org_type, :partners, to: :organisation, allow_nil: true
```
autotelik: why ? enrollment has even less to do with partners than organisation

autotelik:
:ship: from me

app/models/flood_risk_engine/enrollment.rb
reggieb: Because you need to be able to get to partners from enrollment without having to check for organisation all the time.

autotelik: An enrollment does not have partners. Partners are an attribute of Organisation, and only one specific type of Organisation

PR #158, opened 2016-06-15
Fix issues where manual address next page was wrong, and back issue

reggieb:
From [RIP-230](https://eaflood.atlassian.net/browse/RIP-230) comment:

> Registration still misses out contact detail pages after address only if you use an unrecognised postcode (bs99 9xx) and enter the address manually.
> Also if you go back to the address page and select 'I can’t find the address in the list' it errors.

### DIFF
--- a/app/forms/flood_risk_engine/steps/base_address_form.rb
+++ b/app/forms/flood_risk_engine/steps/base_address_form.rb
@@ -32,14 +32,18 @@ module FloodRiskEngine
 
       # read only param for displaying the Postcode in the view
       def postcode
-        return enrollment.address_search.postcode if enrollment.address_search.postcode.present?
-
-        return enrollment.organisation.primary_address.postcode if enrollment.organisation.primary_address.present?
-
-        ""
+        address_search_postcode || primary_address_postcode || ""
       end
 
       protected
+
+      def address_search_postcode
+        enrollment.address_search.try(:postcode)
+      end
+
+      def primary_address_postcode
+        enrollment.organisation.try(:primary_address).try(:postcode)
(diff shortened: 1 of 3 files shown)


app/forms/flood_risk_engine/steps/base_address_form.rb
```diff
@@ -32,14 +32,18 @@ def save
...
-        return enrollment.address_search.postcode if enrollment.address_search.present?
+        address_search_postcode || primary_address_postcode || ""
+      end
 
-        return enrollment.organisation.primary_address.postcode if enrollment.organisation.primary_address.present?
+      protected
 
-        ""
+      def address_search_postcode
+        enrollment.address_search.try(:postcode)
```
autotelik: try is ugly.  your delegate style would be better leading to `organisation.primary_address_postcode`

spec/forms/flood_risk_engine/steps/other_address_form_spec.rb
```diff
@@ -62,6 +62,25 @@ module Steps
           expect(address.addressable_type).to eq("FloodRiskEngine::Organisation")
         end
       end
+
+      describe "#address" do
```
autotelik: :+1:

autotelik:
:ship:

app/forms/flood_risk_engine/steps/base_address_form.rb
reggieb: I think its a case of how often you need to do it. I'll use `try` if it's a one off, and `delegate` if it's used in multiple places.

PR #162, opened 2016-06-16
Fix js error on user_type step

woodpigeon:
The html5 details element must contain a div (wrapping the togglable content) in order to keep the details.polyfill happy - the js error was because this was missing.
Removed details.polyfill from engine application.js and added to the dummy app; this file is included in the FO (from govuk_elements_rails gem) and is part of the 'style and progressively enhance the UI outside of the engine' approach we are taking, so does not belong in the engine.


reggieb:
Looks better than my attempted fix :smile:

:ship: it!

PR #163, opened 2016-06-16
Add errors controller

reggieb:
Adds

- an errors controller
- hooks that will direct standard errors (401 etc.) to the error controller
- hooks in the steps controller to direct some known errors to the errors controller

Also converts journey token to list of tokens, so that you can return to a previous journey if it was in made in the current session (as defined by the cookie)

### DIFF
--- a/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
+++ b/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
@@ -11,6 +11,7 @@ module FloodRiskEngine
     class StepsController < ApplicationController
 
       rescue_from StepError, with: :step_not_found
+      rescue_from JourneyError, ActiveRecord::RecordInvalid, with: :step_not_valid
 
       before_action :check_step_is_valid
       before_action :back_button_cache_buster
@@ -51,16 +52,17 @@ module FloodRiskEngine
       def step_url(options = {})
         enrollment_step_path(enrollment, enrollment.current_step, options)
       end
+      helper_method :step_url
 
       def check_step_is_valid
         check_journey_valid
         return true if step_is_current?
         return step_back if step_back_is_possible?
-        raise StepError, "Requested #{step}, is not permitted when enrollment.step is #{enrollment.current_step}"
+        raise StepError, "Requested :#{step}, is not permitted when enrollment.step is :#{enrollment.current_step}"
       end
 
       def check_journey_valid
-        return true if enrollment.token == cookies.encrypted[:journey_token]
+        return true if journey_tokens.include?(enrollment.token)
         raise(JourneyError, "Journey not started in current browser session")
       end
 
@@ -108,8 +110,13 @@ module FloodRiskEngine
         @enrollment ||= Enrollment.find_by_token!(params[:enrollment_id])
       end
 
-      def step_not_found
-        Rails.logger.info "Step Mismatch: :#{step} requested when enrollment at :#{enrollment.step}"
+      def step_not_valid(exception)
+        Rails.logger.info "STEP NOT VALID: #{exception.message}"
+        redirect_to error_path(:step_not_valid)
+      end
+
+      def step_not_found(exception)
+        Rails.logger.info "STEP NOT FOUND: #{exception.message}"
         redirect_to step_url
       end
 
--- a/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
+++ b/app/controllers/flood_risk_engine/enrollments/steps_controller.rb
@@ -52,7 +52,6 @@ module FloodRiskEngine
       def step_url(options = {})
         enrollment_step_path(enrollment, enrollment.current_step, options)
       end
-      helper_method :step_url
 
       def check_step_is_valid
         check_journey_valid
--- a/app/controllers/flood_risk_engine/errors_controller.rb
+++ b/app/controllers/flood_risk_engine/errors_controller.rb
@@ -12,8 +12,8 @@ module FloodRiskEngine
 
     protected
 
-    def error
-      @error ||= params[:id]
+    def error_code
+      @error_code ||= params[:id]
     end
 
     def template_exists(name)
@@ -28,7 +28,7 @@ module FloodRiskEngine
     end
 
     def template
-      @template ||= template_exists(error) ? error : "generic"
+      @template ||= template_exists(error_code) ? error_code : "generic"
     end
 
     def file_for(name)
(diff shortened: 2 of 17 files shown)


app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -108,8 +110,13 @@ def enrollment
...
-      def step_not_found
-        Rails.logger.info "Step Mismatch: :#{step} requested when enrollment at :#{enrollment.step}"
+      def step_not_valid(exception)
+        Rails.logger.info "STEP NOT VALID: #{exception.message}"
+        redirect_to error_path(:step_not_valid)
+      end
+
+      def step_not_found(exception)
+        Rails.logger.info "STEP NOT FOUND: #{exception.message}"
         redirect_to step_url
```
reggieb: I've left `step_not_found` so that it works as the "pre-errors-controller" system did. That it, it just logs the issue and redirects to the current step. I've left it this way so that it can be compared to the new - "show an error message and let the user decide" way of doing things - so that we can compare and contrast in demos.

reggieb:
As I can hope you can see - the current version has a range of templates that it uses depending on the error. I think these can be rationalised down to one or two. I've left a few so people can see the options available. 

Some of the templates have links, other don't. The generic doesn't - I'm assuming that will appear for fairly serious errors - and giving them a link that will just lead to another error isn't a good idea.

app/controllers/flood_risk_engine/enrollments/steps_controller.rb
```diff
@@ -51,16 +52,17 @@ def step
       def step_url(options = {})
         enrollment_step_path(enrollment, enrollment.current_step, options)
       end
+      helper_method :step_url
```
reggieb: I need to remove this - it was used in an early version, but is no longer needed.

autotelik:
any reason why we're not just dropping in all the DCS stuff  ? their controller seemed alot simpler and we'd get consistency of look and feel out the box - something I thought the designers and tech leads were keen on, and would make the CI team's life easier.

spec/controllers/flood_risk_engine/enrollments/steps_controller_spec.rb
```diff
@@ -5,86 +5,26 @@ module Enrollments
       routes { Engine.routes }
       render_views
       let(:enrollment) { FactoryGirl.create(:enrollment, step: step) }
-
-      context "control of steps" do
```
reggieb: Section removed as it is a duplicate of section in steps_controller_show_spec

spec/dummy/config/environments/development.rb
```diff
@@ -10,7 +10,7 @@
   config.eager_load = false
 
   # Show full error reports and disable caching.
-  config.consider_all_requests_local       = true
+  config.consider_all_requests_local       = false
```
reggieb: Set this to false if you want to see the behaviour in develop. I need to change it back to true in this PR

reggieb:
I've taken this design very much from WEX.

config/routes.rb
```diff
@@ -10,4 +10,8 @@
     )
   end
 
+  resources :errors, only: [:show]
+
+  # See http://patrickperey.com/railscast-053-handling-exceptions/
+  get "(errors)/:id", to: "errors#show"
```
reggieb: This is similar to WEX but I've used `:id` rather than `:status` so that the show action behaves as normal. When rack generates an error it sends it to `errors/<http_status>`, so the status is passed through to the action as the `:id`. That means you don't have to dig around within the exception to get the status as WEX does.

config/routes.rb
```diff
@@ -10,4 +10,8 @@
     )
   end
 
+  resources :errors, only: [:show]
+
```
woodpigeon: Is `resources :errors` needed here? I tried briefly uncommenting it and it still worked

reggieb: Possible not ....

app/controllers/flood_risk_engine/errors_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+        template: file_for(template),
+        locals: { message: exception.try(:message), enrollment: enrollment }
+      )
+    end
+
+    protected
+
+    def error
+      @error ||= params[:id]
+    end
```
woodpigeon: Is @error the error code eg 404? If so error_code might be a clearer name as error could be the exception object or a message

app/controllers/flood_risk_engine/errors_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      return unless journey_token
+      Enrollment.find_by(token: journey_token)
+    end
+
+    def journey_token
+      @journey_token ||= journey_tokens.last
+    end
+
+  end
+end
```
woodpigeon: A minor thought is that the individual small methods used to determine and format the template path could be in a separate class (pseudo code below):
```
class ErrorsController < ApplicationController

    def show
      template = ExceptionTemplate.new(code: error_code, exception: exception)
      render(
        template: template.path
        locals: { message: exception.try(:message), enrollment: enrollment }
      )
    end

    protected

    class ExceptionTemplate
      def initialize(code:, exception:)
        @code = code
        @exception = exception
      end

      def path
         template_path(@code)
      end

      private

      def template_exists(name)
        File.exist? template_path(name)
      end

      def template_path(name)
        File.expand_path(
          "app/views/#{file_for(name)}.html.erb",
          FloodRiskEngine::Engine.root
        )
      end

      def template
        @template ||= template_exists(error) ? error : "generic"
      end

      def file_for(name)
        "flood_risk_engine/errors/error_#{name}"
      end
    end

```

app/controllers/flood_risk_engine/errors_controller.rb
reggieb: Actually, I think `error_status` would be a better name. I'll change it.

woodpigeon:
Looks good. I can't break it. Haven't tired it the BO and FO and works for all the unhandled errors I tried. Two minor comments nothing to stop it :ship:ing when you're happy and squashed if necessary.

config/routes.rb
reggieb: Actually, it adds the `error_path` to routes. However, I can use `as: error` on the get path to provide that functionality

app/controllers/flood_risk_engine/errors_controller.rb
reggieb: I went for error_code - I think you were right in the first place.

app/controllers/flood_risk_engine/errors_controller.rb
reggieb: I'd agree if I wanted to reuse that code, but I don't. It's only here when the controller is called directly out of the rack stack (when rack detects an error) that you have to jump through all these hoops to pick out the right paths to the templates. So other controllers would need anything like `ExceptionTemplate`

PR #164, opened 2016-06-16
RIP-224 Add cookies page

woodpigeon:
- add a cookies page will list out all cookies used in the app.
- rename _privacy_policy.en.yml to privacy_policy.en.yml
- move content in static_pages.en.yml to privacy_policy.en.yml and remove the file
- remove a spec that checks for a privacy policy link as it returns a false positive and test is no longer required


woodpigeon:
👀  - note there is a lot of text on the view that is not in I18n - quite a large task to 'translate' it all and possibly not worthwhile in this instance - any thoughts?

app/views/pages/cookies.html.erb
```diff
@@ -0,0 +1,131 @@
...
+         </tr>
+       </thead>
+       <tbody>
+         <tr>
+           <td>_ga</td>
+           <td>Used to identify unique users and inform referring sites, visitor and session counts</td>
+           <td>2 years</td>
+         </tr>
+       </tbody>
+     </table>
```
reggieb: I wonder if a [description list](http://www.w3schools.com/tags/tag_dl.asp) would be a better way of  displaying this information. More semantic - see https://css-tricks.com/utilizing-the-underused-but-semantically-awesome-definition-list/

reggieb:
For me, it's a consistency thing. I18n translations have been used everywhere else in this project, therefore it should be done here too. 

However, considering time constraints I understand why you might want to get this code in as it is, but if you do that I think you should add a RIP task to move the text into a translation file at a later date. Then if there is time at the end of the project, this can be picked up and done.

autotelik:
I can try to use copykit to convert the HTML to view + locale data if you want .. did pretty good job on privacy

reggieb:
@autotelik That might be the simplest solution.

woodpigeon:
@autotelik if you could run  copykit and send me the output that would be great - even if its not perfect

app/views/pages/privacy_policy.html.erb
```diff
@@ -22,7 +22,7 @@
       <li><%= t('.list_item2') %></li>
     </ul>
 
-    <br>
+    <br/>
```
reggieb: Picky - but closing a `<br>` tag in HTML 5 [is optional](http://stackoverflow.com/questions/1946426/html-5-is-it-br-br-or-br) - with the default being not to.

reggieb:
Well. It's consistent. :ship: it.

app/views/pages/privacy_policy.html.erb
woodpigeon: Good point I didn't know that - I ran htmlhint and it complained so I fixed it. Must be the way we have it configured.

PR #165, opened 2016-06-17
Add partnership details functionality with aid of partners controller

reggieb:
Provides a list of partners added in the previous steps, a link to add another partner, and links to remove individual partners.

Also, only displays a continue button once two partners present.


app/factories/flood_risk_engine/steps/form_object_factory.rb
```diff
@@ -19,23 +19,15 @@ def form_object_class_map
...
           form_name = "FloodRiskEngine::Steps::#{step.to_s.classify}Form"
-          begin
-            form_name.constantize
-          rescue NameError => x
-            Rails.logger.debug(x.backtrace.first)
-            Rails.logger.debug(x.inspect)
-            Rails.logger.debug("Error loading Form class #{form_name} ")
-            nil
-          end
+          form_name.constantize
```
autotelik: why remove this ?

app/views/flood_risk_engine/enrollments/partners/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<p>
+  Please confirm that you wish to remove <%= @partner.full_name %>.
```
autotelik: plain text

reggieb: Good point!

app/views/flood_risk_engine/enrollments/steps/_partnership_details.html.erb
```diff
@@ -1 +1,35 @@
...
+          <% partner.address.parts.each do |address_part| %>
+            <%= address_part %><br>
+          <% end %>
+        </td>
+        <td>
+          <%= link_to(
+                enrollment_partner_path(form.enrollment, partner),
+                class: "link",
+                method: :delete,
+                data: {:confirm => "Are you sure you wish to remove #{partner.full_name}?"}
```
reggieb: And again.

spec/controllers/flood_risk_engine/enrollments/steps_controller/partnership_details_spec.rb
```diff
@@ -0,0 +1,64 @@
+require "rails_helper"
+module FloodRiskEngine
+  module Enrollments
+
+    describe StepsController do
```
autotelik: given that smooth navigation backwards is important to user experience, seems very little rspec testing around what happens at different stages should user click back links or buttons

spec/controllers/flood_risk_engine/enrollments/steps_controller/partnership_details_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+          before do
+            partner2 # ensure 2nd partner exists before page rendered
+            get :show, id: step, enrollment_id: enrollment
+          end
+
+          it "should display partner address" do
+            expect(response.body).to match(address.postcode)
+          end
+
+          it "should display second partner address" do
```
autotelik: no tests related to the display of errors and error anchors ?

app/factories/flood_risk_engine/steps/form_object_factory.rb
reggieb: Because the default error raised when the name can't be converted into a constant is more informative than the custom error. That is, you get an error that tells you precisely what the problem is.

When I started building the form, I gave it the wrong name. I couldn't work out what the problem was until I removed the custom error catching. As the default error was more helpful in debugging the problem, I didn't put the customer error catch back in.

autotelik:
:ship:   once text localised

spec/controllers/flood_risk_engine/enrollments/steps_controller/partnership_details_spec.rb
reggieb: As the same mechanism is used to provide the back step at each step, you shouldn't need to test it on every step.

spec/controllers/flood_risk_engine/enrollments/steps_controller/partnership_details_spec.rb
reggieb: There is no error output on the partnership details page.

spec/controllers/flood_risk_engine/enrollments/steps_controller/partnership_details_spec.rb
autotelik: well experience is showing that making this assumption is dangerous. plus this is not a normal flow is it, got its own controller setting steps etc hasn't it ?

PR #166, opened 2016-06-17
Remove CheckLocation step and contact ea location page

woodpigeon:
This a housekeeping PR.

* Remove the unsued CheckLocation form and spec, and the
redundant `redirection_url_on_location_unchecked` config setting

* Remove the contact ea location page as it is no longer required.
However another contact page (see RIP-147) for the "other"
user type ends in a similar page, so rename
contact_ea_location.html.erb to contact_ea.html.erb and add
the current v3 protoype content. Note that the page correct content
and addition of back link will be addressed in RIP-147.


config/locales/pages/contact_ea.en.yml
```diff
@@ -0,0 +1,12 @@
+---
+en:
+  pages:
+    contact_ea:
+      heading:  Contact the Environment Agency
+      subheading: You chose ‘I don’t know’ for the organisation type
```
reggieb: If the text uses a `'` then the enclosing quotes should be double I think:

> You chose "I don’t know" for the organisation type

config/locales/pages/contact_ea.en.yml
```diff
@@ -0,0 +1,12 @@
...
+      call_to_action: >
+        Please call our helpline to get advice about who might be the organisation or
+        person responsible for your activity.
+      contact_tel: 'Telephone: 03708 506 506'
+      contact_minicom: 'Minicom: 03702 422 549'
+      contact_times: Monday to Friday, 8am to 6pm
```
reggieb: Should this data be coming from (or may be overriden by) environment variables as these details are likely to change.

Also - should use double quotes to be consistent with rest of app.

reggieb:
A couple of minor tweaks suggested. Otherwise :ship: it

PR #172, opened 2016-06-21
Fix/update errors and fix postcode error

reggieb:
Martin has updated the texts of the error pages.

There is also a bug that occurs if a user starts creating a partner and then returns to the partnership page before an address is added to the partner. The result is creation of an incomplete partner. This PR includes a fix for that problem.


woodpigeon:
:ship: it from me

PR #176, opened 2016-06-21
Update activerecord-session_store

woodpigeon:
The previous of activerecord-session_store caused a
deprecation warning `Thread.exclusive is deprecated, use Mutex`.
Upgrading to v 1.0.0 fixes this.


reggieb:
Splendid. :ship: it

PR #177, opened 2016-06-22
Fix error on completing partnership journey, and partners#show styling

reggieb:
An error was being raised when moving to completion for a partnership, because a translation was missing for the email generated.

The `partners#show` page was updates to match the comments in [RIP-94](https://eaflood.atlassian.net/browse/RIP-94)


woodpigeon:
:ship:

PR #179, opened 2016-06-22
Add  submitted date to Export RIP-247

autotelik:

Add submitted_at to Enrollment and set when user ends journey.

### DIFF
--- a/app/models/flood_risk_engine/enrollment.rb
+++ b/app/models/flood_risk_engine/enrollment.rb
@@ -78,6 +78,12 @@ module FloodRiskEngine
       withdrawn: 6        # BO: used to hide anything created in error
     }
 
+    def submit
+      return if submitted_at.present?
+      self.submitted_at = Time.zone.now
+      pending!
+    end
+
     private
 
     # The reference_number in the agreed format "<id with up to 6 zero padded digits>"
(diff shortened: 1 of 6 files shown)


autotelik:
:eyes:

app/models/flood_risk_engine/enrollment.rb
```diff
@@ -78,6 +78,12 @@ def state_machine
       withdrawn: 6        # BO: used to hide anything created in error
     }
 
+    def submit
+      return if submitted_at.present?
+      self.submitted_at = Time.zone.now
+      pending!
```
reggieb: Should the `pending!` method be called on submission even when there is an existing `submitted_at`?

reggieb:
:ship: subject to addressing the `pending!` issue.

app/models/flood_risk_engine/enrollment.rb
autotelik: I don't believe so, this should be a one off, and if pending needs setting in other circumstances can be fired individually - AFAIK this check wanted, to stop some issues found in WEX with states getting reset, when NCCC started editing enrollments

reggieb: OK - Just thought I'd check.

PR #181, opened 2016-06-23
Allow same browser requirement to be configurable.

reggieb:
Adds a configuration option to the engine, that is checked before checking that the same browser is used
through the user journey.

This modification will make it easier to turn off this functionality in the back office app.


autotelik:
:ship:

PR #185, opened 2016-06-27
Add comments model and associate with enrollment exemptions

reggieb:
Comments are needed in the back office, but need to be associated with database objects defined in the engine, so defining model here.

### DIFF
--- a/app/models/flood_risk_engine/comment.rb
+++ b/app/models/flood_risk_engine/comment.rb
@@ -1,0 +1,6 @@
+module FloodRiskEngine
+  class Comment < ActiveRecord::Base
+    belongs_to :commentable, polymorphic: true
+    belongs_to(:user) if defined?(User)
+  end
+end
--- a/app/helpers/flood_risk_engine/application_helper.rb
+++ b/app/helpers/flood_risk_engine/application_helper.rb
@@ -36,7 +36,7 @@ module FloodRiskEngine
       if form && form.errors[attribute].any?
         classes << "error"
         content = content_tag(:span,
-                              form.errors[attribute].first.to_s.html_safe,
+                              form.errors[attribute].first,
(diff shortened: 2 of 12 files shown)


app/models/flood_risk_engine/comment.rb
```diff
@@ -0,0 +1,6 @@
+module FloodRiskEngine
+  class Comment < ActiveRecord::Base
+    belongs_to :commentable, polymorphic: true
+    belongs_to(:user) if defined?(User)
```
autotelik: why is the if needed and is this a good way to modulate access  ? would't any access to the behaviours mixed in, also have to be guarded in a similar way ?

reggieb: The `if` is there, because the User model is defined in the BO only. So in other environments, the `belongs_to` needs to be disabled.

Also the `user` method is not being used to control access, but to record who added the comment.

autotelik: this is not the way to do it then. the belongs_to(:user)  should be defined in the BO app only

reggieb: If you move `Comment` to BO you have problems with the joins from objects using the comments, such as `EnrollmentExemption` (that's where I started). 

It looks to me like we should move all models to the engine - because of the issues around database joins being defined across the engine and host app.

autotelik: app/decorators/models/flood_risk_engine

reggieb: Yes, exactly the sort of ugliness I'm trying to avoid.

autotelik: this is a good read 

http://edgeguides.rubyonrails.org/engines.html

PR #186, opened 2016-06-27
RIP-163-164 Add a shared application version page

woodpigeon:
This can be accessed from the FO or BO, and displays
* the application name, pulled from config or defaulting
  to the Application class name
* The application version, as defined by the host app
  in Application::VERSION
* the currently deployed git sha with a link to that commit,
  using the git repo url defined in a config setting, but
  defaulting to a github repo derived from the Application
  class name


reggieb:
Looks fine. :ship: it.

PR #194, opened 2016-07-01
Fields for decision maker and date - RIP-247

autotelik:

### DIFF
--- a/app/models/flood_risk_engine/enrollment_exemption.rb
+++ b/app/models/flood_risk_engine/enrollment_exemption.rb
@@ -20,5 +20,25 @@ module FloodRiskEngine
       includes(:exemption).where(flood_risk_engine_exemptions: { code: Exemption::LONG_DREDGING_CODES }).any?
     end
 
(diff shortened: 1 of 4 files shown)


autotelik:
:eyes:

app/models/flood_risk_engine/enrollment_exemption.rb
```diff
@@ -20,5 +20,21 @@ def self.include_long_dredging?
...
+    def decision_at
+      return if comments.empty?
+      comments.order(:created_at).last.created_at
+    end
+
+    def decision_user_id
+      return if comments.empty?
+      comments.order(:created_at).last.user_id
+    end
+
```
reggieb: I'd suggest adding a latest comment method that could then be used in all three methods:
```ruby
def latest_comment
  comments.order(:created_at).last
end
```

reggieb:
I think there should be some tests of these methods.

autotelik:
scope and tests added

PR #196, opened 2016-07-01
RIP-249 Display dredging and site info in rows

lewispb:
https://eaflood.atlassian.net/browse/RIP-249

This only displays dredging length on the 'check your answers' page
if one of the enrollment exemptions is a long dredging exemption.

It always displays the location or site description on the 'check your
answers' page.

**Question for reviewer:** RowBuilder is now over 100 lines so I had to *disable the rubocop check* for class length here. Is that acceptable? If not, how would you approach this?

### DIFF
--- a/app/presenters/flood_risk_engine/tabular_enrollment_detail/row_builder.rb
+++ b/app/presenters/flood_risk_engine/tabular_enrollment_detail/row_builder.rb
@@ -1,3 +1,4 @@
+# rubocop:disable ClassLength
 module FloodRiskEngine
   module TabularEnrollmentDetail
     class RowBuilder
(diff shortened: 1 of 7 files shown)


app/presenters/flood_risk_engine/tabular_enrollment_detail/row_builder.rb
```diff
@@ -42,6 +43,22 @@ def exemption_row(exemption)
...
+        return unless enrollment_presenter.exemption_location
+        build_row name: :location_description,
+                  value: enrollment_presenter.exemption_location.description,
+                  step: :grid_reference
+      end
+
+      def location_dredging_length_row
+        return unless enrollment.enrollment_exemptions.include_long_dredging?
+        dredging_length = enrollment_presenter.exemption_location.dredging_length
+        formatted_dredging_length = ActiveSupport::NumberHelper.number_to_human(dredging_length, units: :distance)
```
woodpigeon: Its only 11am and I've learnt something new already! Nice

lewispb:
@woodpigeon Thanks for review. Did you see my note above about RowBuilder now being over 100 lines? Is that an issue?

PR #201, opened 2016-07-05
Add enrollment.submitted? and enrollment_exemption.status_one_of?

reggieb:
Provides a better way of determining when an enrollment has been submitted, and matching enrollment exemption status with a set of alternatives.

Also some tidying up (see comments below)

### DIFF
--- a/app/models/flood_risk_engine/enrollment.rb
+++ b/app/models/flood_risk_engine/enrollment.rb
@@ -34,17 +34,14 @@ module FloodRiskEngine
 
     has_one :address_search
 
-    serialize :step_history, Array
-
     before_validation :preserve_current_step
+    after_create :generate_and_save_reference_number
 
     # Have to use `validate` as `validates` is called on page load, before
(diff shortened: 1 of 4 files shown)


app/models/flood_risk_engine/enrollment.rb
```diff
@@ -34,18 +34,15 @@ class Enrollment < ActiveRecord::Base
 
     has_one :address_search
 
-    serialize :step_history, Array
-
     before_validation :preserve_current_step
+    after_create :generate_and_save_reference_number
```
reggieb: Moved this so that the callbacks are together

app/models/flood_risk_engine/enrollment.rb
```diff
@@ -34,18 +34,15 @@ class Enrollment < ActiveRecord::Base
 
     has_one :address_search
 
-    serialize :step_history, Array
```
reggieb: step history isn't used anywhere.

PR #202, opened 2016-07-05
RIP-295 Update submitted email text

lewispb:
https://eaflood.atlassian.net/browse/RIP-295

Minor tweaks to headings on the submitted confirmation email.

### DIFF
--- a/config/locales/flood_risk_engine/enrollment_mailer.en.yml
+++ b/config/locales/flood_risk_engine/enrollment_mailer.en.yml
@@ -3,9 +3,9 @@ en:
   flood_risk_engine:
     enrollment_mailer:
       submitted:
-        subject: Registration submitted
+        subject: Flood risk activity exemption submitted
+        heading_html: 'Registration submitted: <br /> please wait for us to confirm it'
         reference_number_heading: Your reference is %{reference_number}
-        please_wait: Please wait for us to confirm it before you start work.
         exemption_heading: Exemption submitted
         text_heading: 'Registration submitted: please wait for us to confirm it'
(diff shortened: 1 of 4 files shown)


config/locales/flood_risk_engine/enrollment_mailer.en.yml
```diff
@@ -3,9 +3,9 @@ en:
   flood_risk_engine:
     enrollment_mailer:
       submitted:
-        subject: Registration submitted
+        subject: Flood risk activity exemption submitted
+        heading_html: 'Registration submitted: <br /> please wait for us to confirm it'
```
reggieb: I don't think you should be passing markup within a translation. I'd suggest breaking this down into elements and adding the break in the view code.

config/locales/flood_risk_engine/enrollment_mailer.en.yml
```diff
@@ -3,9 +3,9 @@ en:
...
         reference_number_heading: Your reference is %{reference_number}
-        please_wait: Please wait for us to confirm it before you start work.
         exemption_heading: Exemption submitted
         text_heading: 'Registration submitted: please wait for us to confirm it'
```
reggieb: Having text and html versions of the same text feels wrong.

lewispb: Good spot, I agree and I've removed this now.

config/locales/flood_risk_engine/enrollment_mailer.en.yml
lewispb: Ok, I see what you mean. Done.

PR #203, opened 2016-07-05
Database fields to hold accept reject decision data

autotelik:

Clarity from NCCC :

· Decision user – This will be the person that has screened the registration and made the decision as to whether accept or reject.

· Decision date – this is the date we accept or reject the registration.

We need the decision date aswell as the submitted date to always remain the same, this is so we can compare how we are against our service level agreement. The decision maker we would want to show as the originally person that either accepted or rejected.

### DIFF
--- a/app/models/flood_risk_engine/enrollment_exemption.rb
+++ b/app/models/flood_risk_engine/enrollment_exemption.rb
@@ -20,30 +20,6 @@ module FloodRiskEngine
       includes(:exemption).where(flood_risk_engine_exemptions: { code: Exemption::LONG_DREDGING_CODES }).any?
     end
 
-    def accept_reject_decision_user
-      User.find_by_id(accept_reject_decision_user_id).try(:email)
-    end
-
-    def latest_comment_at_and_user_id
-      comments.order(:created_at).pluck(:created_at, :user_id).last
-    end
-
-    def decision_at_and_user
-      return [nil, nil] if comments.empty?
-      latest = latest_comment_at_and_user_id
-      [latest.first, User.find(latest.last).try(:email)]
-    end
-
-    def decision_at
-      return if comments.empty?
-      latest_comment_at_and_user_id.first
-    end
-
-    def decision_user_id
-      return if comments.empty?
-      latest_comment_at_and_user_id.last
-    end
-
     def status_one_of?(*statuses)
       statuses.collect(&:to_s).include? status
     end
--- a/app/models/flood_risk_engine/enrollment_exemption.rb
+++ b/app/models/flood_risk_engine/enrollment_exemption.rb
@@ -20,24 +20,28 @@ module FloodRiskEngine
       includes(:exemption).where(flood_risk_engine_exemptions: { code: Exemption::LONG_DREDGING_CODES }).any?
     end
 
-    def latest_decision
+    def accept_reject_decision_user
+      User.find_by_id(accept_reject_decision_user_id).try(:email)
+    end
+
+    def latest_comment_at_and_user_id
       comments.order(:created_at).pluck(:created_at, :user_id).last
     end
 
     def decision_at_and_user
       return [nil, nil] if comments.empty?
-      latest = latest_decision
+      latest = latest_comment_at_and_user_id
       [latest.first, User.find(latest.last).try(:email)]
     end
 
     def decision_at
       return if comments.empty?
-      latest_decision.first
+      latest_comment_at_and_user_id.first
     end
 
     def decision_user_id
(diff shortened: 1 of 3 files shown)


autotelik:
:eyes:

app/models/flood_risk_engine/enrollment_exemption.rb
```diff
@@ -20,24 +20,28 @@ def self.include_long_dredging?
       includes(:exemption).where(flood_risk_engine_exemptions: { code: Exemption::LONG_DREDGING_CODES }).any?
     end
 
-    def latest_decision
+    def accept_reject_decision_user
+      User.find_by_id(accept_reject_decision_user_id).try(:email)
```
lewispb: Is it worth setting this up as an association?

`belongs_to :accept_reject_decision_user, class_name: 'User'` or something like that.

Then this method would be called `accept_reject_decision_user_email`

app/models/flood_risk_engine/enrollment_exemption.rb
```diff
@@ -20,24 +20,28 @@ def self.include_long_dredging?
...
+    end
+
+    def latest_comment_at_and_user_id
       comments.order(:created_at).pluck(:created_at, :user_id).last
     end
 
     def decision_at_and_user
       return [nil, nil] if comments.empty?
-      latest = latest_decision
+      latest = latest_comment_at_and_user_id
```
lewispb: Rather than just plucking two attributes and using them, do you think the code would read better if you just retrieved the whole comment and then used the attributes you wanted within the other methods?

app/models/flood_risk_engine/enrollment_exemption.rb
autotelik: yeah, good spot, don't know what I was thinking !! :-)

woodpigeon: As this is the first time we are referencing the User model in the engine (User model currently defined in th BO) - am just having a think if there is an alternative to avoid coupling the engine with the BO..

app/models/flood_risk_engine/enrollment_exemption.rb
autotelik: comments could contain largish strings so I think the efficiency pay off, worth possible readability issues

app/models/flood_risk_engine/enrollment_exemption.rb
reggieb: `User` doesn't exist in the Engine. If you do this we get into all sorts of dependency issues. This either needs to be done in a presenter in the BO, or we move the `User` object out of the BO and into the Engine.

It is very important that any dependancy only goes one way. The engine MUST NEVER depend on the behaviour of something defined in one of it's host applications. That way madness lies.

Sorry I didn't pick this up yesterday when I had a quick look at the code.

autotelik: Yes, I would have put it in a decorator in BO but couldn't be bothered dealing with the ensuing comments.  I guess that engine is never going be used for another app, so from pragmatic perspective coupling not really an issue

woodpigeon: Yes I think @reggieb is right - as the methods are for presentation only they could go into the BO EnrollmentExemptionPresenter for instance

app/models/flood_risk_engine/enrollment_exemption.rb
lewispb: use `select` instead of `pluck`?

`comments.order(:created_at).select(:created_at, :user_id).last`

app/models/flood_risk_engine/enrollment_exemption.rb
reggieb: It doesn't need one of your "decorators". You are gathering data together for output to a page - this is the perfect job for a presenter!

autotelik: belongs_to :accept_reject_decision_user, class_name: 'User'

app/models/flood_risk_engine/enrollment_exemption.rb
autotelik: removed completely

autotelik:
Moved everything into BO

app/models/flood_risk_engine/enrollment_exemption.rb
reggieb: In hindsight, in BO I should have done:
```ruby
class Comment < FloodRiskEngine::Comment
  belongs_to :user
end
```

PR #205, opened 2016-07-06
RIP-299 Update check answers caption

lewispb:
https://eaflood.atlassian.net/browse/RIP-299

Update table caption wording, as supplied.


reggieb:
Horrible grammar - but it's the texted specified - so :ship:

reggieb:
Hold fire - Martin's coming up with a rewording.

reggieb:
Martin has updated the story. The text should be:

> The information you've given us

PR #208, opened 2016-07-06
Add system to create compound reference number for enrollments

reggieb:
[RIP-244](https://eaflood.atlassian.net/browse/RIP-244)

Adds prefix and offset to reference number.

### DIFF
--- a/app/services/flood_risk_engine/reference_number.rb
+++ b/app/services/flood_risk_engine/reference_number.rb
@@ -1,0 +1,58 @@
+module FloodRiskEngine
+  class ReferenceNumber
+
+    def self.with(params)
+      new(params).number
+    end
+
+    attr_reader :seed, :offset, :minimum_length, :padding, :prefix
+    def initialize(
+        seed: nil,
+        offset: nil,
+        minimum_length: nil,
+        padding: nil,
+        prefix: nil
+    )
+      @seed = seed
+      @offset = offset
+      @minimum_length = minimum_length
+      @padding = padding
+      @prefix = prefix
+      check_valid
+    end
+
+    def number
+      @number ||= [prefix, body].compact.join
+    end
+
+    private
+
+    # Main part of reference number without the prefix
+    def body
+      return increment unless minimum_length
+      padded_increment
+    end
+
+    # The part of the reference number incremented to make number unique
+    def increment
+      return seed.to_s unless offset
+      (offset + seed).to_s
+    end
+
+    def padded_increment
+      increment.rjust(body_length, padding)
+    end
+
+    # Note that minimum_length defines total length of number, so
+    # needs to take into account prefix length if present
+    def body_length
+      return minimum_length unless prefix && minimum_length
+      minimum_length - prefix.length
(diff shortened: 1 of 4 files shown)


app/services/flood_risk_engine/reference_number.rb
```diff
@@ -0,0 +1,58 @@
+module FloodRiskEngine
```
autotelik: this class makes it hard to follow whats actually being built and returned. perhaps over engineered giving its really just formatting a string

autotelik:
:ship:  from me

app/services/flood_risk_engine/reference_number.rb
reggieb: I know what you mean, but my starting point was how it would be used:

```ruby
number = ReferenceNumber.with(
  prefix: "EXFRA",
  seed: id,
  offset: 10_000,
  padding: "0",
  minimum_length: 11
)
```

And as I thought there was a reasonable chanced that any of these options could be changed or dropped, I needed to make it flexible enough to handle that.

Also although it is only formatting a string, there is a lot of complication in there - adding the prefix, padding, and offset.

woodpigeon:
A nice solution (I like the starting point though the implementation a tad verbose as @autotelik says - though you could use virtus (already in the engine) - to remove initialize - though its a bit of magic)). 

Couple of points
* The reference number needs to be associated with the EnrollmentExemption not the Enrollment - each exemption gets a unique reference. We may need to chat about this if moving it onto the EE is not straightforward. I know the confirmation page pumps out the reference number which in WEX was the enrollment reference. If we support multiple exemptions per enrollment in the future we'd have to list out multiple reference numbers there I guess. 
* My solution might have been different - maybe something along the lines of overriding the `rake db:reset` task maybe eg (not tested!)
```Rake::Task['db:create'].enhance do
  # invoke a task which calls 
  # ActiveRecord::Base.connection.execute("ALTER SEQUENCE
  #   flood_risk_engine_enrollment_exemptions_id_seq START with 10000 RESTART;")
end
 then have one method on EE to format the id into a reference_number and save in a callback like you do. Haven't thought it through though if honest!

reggieb:
@woodpigeon The more I think about this, the more I don't like the idea of modifying the `db:create`, or otherwise automate the setting of the id sequence within the database. My three main issue are:

1. It separates the responsibility for defining the reference number: That is, part of the way reference number behaves will be in the  ReferenceNumber code, and part in rake command.
2. I feel nervous in how it will behave in situations like disaster recovery and database migration. That is if the app is moved to another database, and existing data is then restored to the newly created table.
3. How this mechanism will behave if the start number needs to be changed again.

Note that points 2 and 3 come from the fact that if you set the sequence start at a point before existing IDs the system will break when the database tries to set an ID that is already being used.

Therefore the key thing is that we need to be aware that the current system uses the ID, and that adding enrollments to the production system will offset the start sequence. So therefore we will need to reset (drop and then recreate) the production database if pre-release test are needed and we need the first live ID to be 010001.

PR #210, opened 2016-07-06
RIP-270 Allow ampersands in organisation name

lewispb:
https://eaflood.atlassian.net/browse/RIP-270

"Canal & River Trust" is correct but generates an error. Need & to be allowed.
Applied this rule to Partnerships and Local Authorities in this commit.

### DIFF
--- a/app/forms/flood_risk_engine/steps/local_authority_form.rb
+++ b/app/forms/flood_risk_engine/steps/local_authority_form.rb
@@ -19,7 +19,10 @@ module FloodRiskEngine
 
       validates :name, presence: { message: t(".errors.name.blank") }
 
-      validates :name, 'flood_risk_engine/text_field_content': true, allow_blank: true
+      validates :name, 'EA::Validators::CompaniesHouseName': {
+        message: t(".errors.name.invalid"),
+        allow_blank: true
+      }
 
       validates :name, length: {
         maximum: LocalAuthorityForm.name_max_length,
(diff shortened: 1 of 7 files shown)


app/forms/flood_risk_engine/steps/partnership_form.rb
```diff
@@ -29,7 +29,7 @@ def partner_count
         presence: {
           message: t(".errors.full_name.blank")
         },
-        "flood_risk_engine/text_field_content" => {
+        "flood_risk_engine/organisation_name_content" => {
```
woodpigeon: I think partnerships are made of people names rather so this should use same validator as Individual (text_field_content)

woodpigeon:
Looks good @lewispb apart from comment on partnerships, but I need to check as it might be easier to just reuse the CompaniesHouseName validator (which will be moved to a separate gem at some point hence its under the EA namespace) as that allows ampersands and more, (e.g. * etc as some companies - and organisations - have unusual characters in their names). So hang on to this one for now and I'll get back to you.

lewispb:
@woodpigeon Ah ok that makes sense thanks Tim. I've reverted the partnership change anyways.

woodpigeon:
@lewispb Checked in with Martin about this and the outcome is that we just need to change
*LocalAuthority* and *Other* name validation to use the CompaniesHouseNameValidator and its associated validation message - sound OK?

woodpigeon:
:ship: looks good

PR #213, opened 2016-07-07
Feature/rip 244 configure exemption number format

reggieb:
Refactor reference number to use a model to store and generate each number in isolation to the enrollment and/or exemption. In particular, this addresses the problem of incomplete enrollments leading to missing numbers in the sequence of reference numbers.


app/models/flood_risk_engine/reference_number.rb
```diff
@@ -0,0 +1,28 @@
+module FloodRiskEngine
+  class ReferenceNumber < ActiveRecord::Base
+    PREFIX = "EXFRA".freeze
+    PADDING = "0".freeze
+    OFFSET = 10_000
```
woodpigeon: According to original story and a meeting today about go live tasks, the expectation is this is 1000 not 10000

app/models/flood_risk_engine/enrollment.rb
```diff
@@ -68,6 +72,8 @@ def state_machine
     def submit
       return if submitted_at.present?
       self.submitted_at = Time.zone.now
+      self.reference_number = ReferenceNumber.create
+      save
```
woodpigeon: Does it need a transaction around it? Ref number might get created ok no line 75 but save on line 76 fails for some reason. Unlikely I know but would prevent gaps in the sequence

woodpigeon:
Added a few small comments - I think the 1000 one is important, transaction one less so. Happy to :ship: when offset adjusted.

app/models/flood_risk_engine/reference_number.rb
reggieb: If you look at the original spec., the offset was given by Polly as 10,000. Not problem changing it to 1000.

PR #214, opened 2016-07-08
RIP-277 Add area lookup

woodpigeon:
* Add ea-area_lookup gem and an initialiser to configure it
* Add WaterBoundaryArea model
* Kick off a background job in the grid ref form #save to fetch
  the area from an api, create it in the db if it does not exist
  and assign it to the location


app/jobs/flood_risk_engine/update_water_boundary_area_job.rb
```diff
@@ -0,0 +1,52 @@
...
+        area.area_name = api_result.area_name
+        area.area_id = api_result.area_id
+      end
+    end
+
+    def inside_england?(api_result)
+      api_result.present? &&
+        api_result.respond_to?(:long_name) &&
+        api_result.long_name.present?
+    end
```
reggieb: I'd suggest replacing the three calls with `!!api_result.try(:long_name)`

spec/models/flood_risk_engine/location_spec.rb
```diff
@@ -1,8 +1,9 @@
 require "rails_helper"
 
 module FloodRiskEngine
-  RSpec.describe Location, type: :model do
+  RSpec.describe Location, type: :model, focusd: true do
```
reggieb: Is `focusd` a typo?

reggieb:
A couple of minor tweaks - but once addressed, :ship:

app/jobs/flood_risk_engine/update_water_boundary_area_job.rb
reggieb: Or even just `api_return.try(:long_name)`

app/jobs/flood_risk_engine/update_water_boundary_area_job.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    def outside_england_admin_area
+      WaterBoundaryArea.find_or_create_by(code: "outside_england") do |area|
+        area.long_name = area.short_name = area.area_name = "Outside England"
```
autotelik: given requires knowledge of 4 methods .. code maybe better within a method on WaterBoundaryArea  e.g def  build_outside_england

app/jobs/flood_risk_engine/update_water_boundary_area_job.rb
```diff
@@ -0,0 +1,52 @@
...
+    def lookup_admin_area_using_using_external_api(location)
+      coords = EA::AreaLookup::Coordinates.new(easting: location.easting,
+                                               northing: location.northing)
+      hash = EA::AreaLookup.find_by_coordinates(coords)
+      OpenStruct.new(hash)
+    end
+
+    def find_or_create_admin_area_from_api_result(api_result)
+      WaterBoundaryArea.find_or_create_by(code: api_result.code) do |area|
+        area.long_name = api_result.long_name
```
autotelik: all the setter methods perhaps better within single method on WaterBoundaryArea  e.g def  build_from_api_result(api_result)

app/jobs/flood_risk_engine/update_water_boundary_area_job.rb
```diff
@@ -0,0 +1,44 @@
...
+      hash = EA::AreaLookup.find_by_coordinates(coords)
+      OpenStruct.new(hash)
+    end
+
+    def find_or_create_area_from_api_result(api_result)
+      WaterBoundaryArea.find_or_create_by(code: api_result.code) do |area|
+        area.long_name = api_result.long_name
+        area.short_name = api_result.short_name
+        area.area_name = api_result.area_name
+        area.area_id = api_result.area_id
```
reggieb: If `lookup_admin_area_using_using_external_api` returns a hash rather than an OpenStruct, I think you'll be able to pass the hash into `attributes` here, and pass them into the WaterBoundaryArea that way - in one line:
```
area.attributes(api_result)
```

PR #223, opened 2016-07-12
RIP-210 Fix multiple partner search results

lewispb:
https://eaflood.atlassian.net/browse/RIP-210

Put in a solution to the multiple partner search result issue by
creating a search content column on the organisation model.

Then we can search that column in the BO without using an outer join
which was returning multiple enrolments.


app/models/flood_risk_engine/organisation.rb
```diff
@@ -14,10 +18,23 @@ class Organisation < ActiveRecord::Base
       unknown: 6
     }
 
-    has_one :primary_address, -> { where(address_type: Address.address_types["primary"]) },
```
lewispb: Rearranged to group associations together

woodpigeon:
Looks good - :ship: when happy

PR #225, opened 2016-07-12
Assisted Mode - RIP-218

autotelik:
https://eaflood.atlassian.net/browse/RIP-218

Set Fully Assisted when a BO user present at the Exemption state

### DIFF
--- a/app/models/flood_risk_engine/enrollment_exemption.rb
+++ b/app/models/flood_risk_engine/enrollment_exemption.rb
@@ -22,6 +22,13 @@ module FloodRiskEngine
       operator_failings: 1
     }
 
+    enum assistance_mode: {
+      unassisted: 0,
+      fully_assisted: 1,
+      partial: 2,
+      admin_intervention: 3
+    }
+
     def self.include_long_dredging?
       includes(:exemption).where(flood_risk_engine_exemptions: { code: Exemption::LONG_DREDGING_CODES }).any?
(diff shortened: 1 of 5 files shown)


app/models/flood_risk_engine/enrollment_exemption.rb
```diff
@@ -22,6 +22,13 @@ class EnrollmentExemption < ActiveRecord::Base
       operator_failings: 1
     }
 
+    enum assistance_mode: {
+      unassisted: 0,
+      fully: 1,
+      partial: 2,
+      admin_intervention: 3
+    }
+
```
woodpigeon: Just wondered if the `fully` and `partial` enum values were e.g. `fully_assisted` and `partially_assisted` then calling `enrollment_exemption.fully_assisted!` would have more meaning than `enrollment_exemption.fully!`

autotelik: yeah been umming and ahhhing whether  to change them, but just went with WEX values n the end for simplicity ..only used in couple of places

woodpigeon:
:ship:

app/models/flood_risk_engine/enrollment_exemption.rb
autotelik: I'm gonna change fully to fully_assisted .. does read much better

PR #230, opened 2016-07-13
Area lookup should find water management area not administrative area

woodpigeon:
- Find the water management area not the the admin area as we
are currently doing
- Rename the WaterBoundaryArea model and relationships to
WaterManagementArea to align with business terms.


reggieb:
Looks fine :ship: :shower: :shaved_ice:

PR #239, opened 2016-07-18
Improved state machine documentation

reggieb:
Adds a README specifically to describe how the state machine is defined. 

Also adds some comments that separate the state machines rules into logical groups.


woodpigeon:
Looks great. Do you think its worth adding a link from the main Readme to this one, so its more easily discovered?

reggieb:
I've a link to the state machine readme, into the main README.

PR #241, opened 2016-07-22
Feature/rip 143 check answers change link

reggieb:
Enables a user to step back to the appropriate location when in review. That is, a link on the review page takes the user to the page where the attribute can be updated. Then on update, the user is returned to the review page.

### DIFF
--- a/app/presenters/flood_risk_engine/tabular_enrollment_detail/row_builder.rb
+++ b/app/presenters/flood_risk_engine/tabular_enrollment_detail/row_builder.rb
@@ -62,13 +62,13 @@ module FloodRiskEngine
       def organisation_name_row
         build_row name: :organisation_name,
                   value: enrollment_presenter.organisation_name,
-                  step: organisation_type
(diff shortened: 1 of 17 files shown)


app/presenters/flood_risk_engine/tabular_enrollment_detail/row_builder.rb
```diff
@@ -62,13 +62,13 @@ def location_dredging_length_row
...
                   value: enrollment_presenter.organisation_name,
-                  step: organisation_type
+                  step: enrollment.org_type
       end
 
       def organisation_address_row
         build_row name: :organisation_address,
                   value: enrollment_presenter.organisation_address,
-                  step: "#{organisation_type}_address"
+                  step: "#{enrollment.org_type}_address"
```
reggieb: `organisation_type` returns a humanised form of the organisation name, but need the symbolised version for links, and that's returned by `enrollment.org_type`

PR #243, opened 2016-11-15
ITSM-1179689 Fix issue with firefox details / summary function

lewispb:
On the grid reference page, using Firefox, it is not possible to enter a NGR because clicking the input box expands or contracts the summary beneath. Clearing the floated element above fixes this.


TThurston:
Approved

PR #245, opened 2016-11-30
Replace before_commit with individual CI checks

Cruikshanks:
[before_commit](https://github.com/EnvironmentAgency/before_commit) was an attempt to have a central gem which could be included in every Ruby based project. The gem would manage the checks we wished to perform in each, thereby ensuring consistency and quality across the projects.
    
In practise we have found when it works, it works great however when it doesn't, it can be very confusing to try and resolve what is actually failing (the project, the check, or **before_commit** itself). We also frequently encountered issues with trying to essentially enforce a centrally controlled workflow on differing development environments.

We have agreed instead to move to ensuring the checks are run independently, and only on the relevant CI server. It will be the responsibility of devs to run these same checks locally, in the way appropriate to their environment, to ensure they don't '**break the build**'.

This change covers removing the dependency on **before_commit** and instead moving to the new agreed model. This involved

- removing the .git-hooks and [Overcommit](https://github.com/brigade/overcommit) (the gem **before_commit** is built on) config and gemfiles from the project
- removing them as dependencies in the project
- removing the require statement and loading of Rake tasks from the `Rakefile`


Cruikshanks:
This is almost there @TThurston @lewispb . Having put the [i18n-tasks](https://github.com/glebm/i18n-tasks) check back in, I'm now getting an error again

```bash
i18n-tasks: Error scanning app/forms/flood_risk_engine/enrollments/address_form.rb: Cannot resolve relative key ".errors.premises.blank".
```

At this time (and hour!) nothing is immediately jumping out to me why this fails and other keys in that view seem fine. I think if we get that error nailed then this will be ready to go into *review*.

It should be flagged though we are still 2 checks shy of what we supposedly had before. We're missing **Scss** and **JavaScript** lint checks.

- The previous [JSHint](https://github.com/EnvironmentAgency/before_commit/blob/master/source/.overcommit.yml#L38) check has been left out as we have now agreed [StandardJS](https://github.com/feross/standard) for all `.js` files. Feel it would be better to implement that, but it can be done after this in a separate PR.
- The previous [ScssLint](https://github.com/EnvironmentAgency/before_commit/blob/master/source/.overcommit.yml#L43) check has been left out. To perform the check you need to install an additional gem, and upon checking its GitHub page found a notification recommending that we [consider alternatives](https://github.com/brigade/scss-lint#notice-consider-other-tools-before-adopting-scss-lint) first. So much like the JavaScript lint check, felt it would be better to complete this PR and tackle that check separately.

If I get a chance at any point I'll keep digging into that locale error, else if you can spare a mo I'd appreciate someone having a look.

Finally if we can get these changes completed and in, it should be much simpler to make the same changes to the remaining 2 repos to get them passing (and the FRAE service building again).

Cruikshanks:
@TThurston @lewispb Ignore my last comment about the **i18n-tasks** error. I've managed to find a fix so the tasks are now running correctly.

The only issue remaining is that its reporting over 270 keys missing, but a closer inspection finds they are being used. The problem comes down to the syntax used and **i18n-tasks** failing to match them up to the right keys in the locale files. A bit like sorting out the Scss and JS linting, this feels like a task in and of itself, so for now I've commented out the command in the `.travis.yml` config.

TThurston:
Everything looks good to me.  Also discussed with @kennyevil - approval also received.  Will squash and merge 👍

PR #249, opened 2016-12-16
Add OS places notice and T&C's to service

Cruikshanks:
It's been highlighted that the service is missing a required notice and link to T&C's related to using OS Places and its data within the service.

Specifically we have to add the notice and link on the page where postcode lookup is used. This changes updates the view with the notice text, and adds a new OS Places T&C's page which is accessed from a link in the notice.

### DIFF
--- a/app/views/flood_risk_engine/shared/_postcode.html.erb
+++ b/app/views/flood_risk_engine/shared/_postcode.html.erb
@@ -33,12 +33,12 @@
 <p class="text font-xsmall" id="os-notice">
   <%=
     t(
-      ".os_notice.text",
+      ".os_notice.text_html",
       link: link_to(
         t(".os_notice.url_text"),
         enrollment_page_path(form.enrollment, id: "os_places_terms"),
         target: "_blank"
       )
-    ).html_safe
+    )
   %>
 </p>
--- a/app/views/flood_risk_engine/shared/_postcode.html.erb
+++ b/app/views/flood_risk_engine/shared/_postcode.html.erb
@@ -22,5 +22,23 @@
       %>
     </p>
   <% end  %>
-
+  <p class="text" id="os-notice">
+    <%=
+      t(
+        ".os_notice.text",
+        link1: link_to(
+          t(".os_notice.url1_text"),
+          t('.os_notice.url1'),
+          rel: "external",
+          target: "_blank"
+        ),
+        link2: link_to(
+          t(".os_notice.url2_text"),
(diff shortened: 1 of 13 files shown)


Cruikshanks:
@TThurston @lewispb . The work is essentially done so the changes cover both new content, the amended postcode page, and the translations and tests that go with it.

The screenshots below demonstrate how they currently look when mounted in the [front office](https://github.com/EnvironmentAgency/flood-risk-front-office).

screen shot 2016-12-18 at 23 38 30

screen shot 2016-12-18 at 23 37 24

Before we complete the changes though I just want to run them by @glancyea and @timstone if possible. We've been asked to reproduce [LTFR](https://flood-warning-information.service.gov.uk/long-term-flood-risk) but we've done things in FRAE, like state when a link will open a new window, that means we can't do that exactly.

Plus where the terms are numbered, I believe from an accessibility point of view we should be putting them in a `<ol>` rather than just `<p>`'s with the numbering manually done. It's things like that I want to check before I mark this **ready for review**.

Cruikshanks:
Have now implemented the feedback I got from @timstone and @glancyea . You can see screenshots of the new versions below.

screen shot 2016-12-19 at 12 26 55

screen shot 2016-12-19 at 12 27 33

On this basis I think the code is *ready for review*. I'm going to ask them to do a final double check and also run it past our contact Nick in DMMI, but am still pushing it due to the time pressures and the hope there won't be any further changes.

app/views/flood_risk_engine/shared/_postcode.html.erb
```diff
@@ -24,3 +24,21 @@
...
+<p class="text font-xsmall" id="os-notice">
+  <%=
+    t(
+      ".os_notice.text",
+      link: link_to(
+        t(".os_notice.url_text"),
+        enrollment_page_path(form.enrollment, id: "os_places_terms"),
+        target: "_blank"
+      )
+    ).html_safe
```
lewispb: Rather than use `html_safe` it's worth noting you can use the i18n key name (end the key name with `_html`) to determine the ouptut of html in i18n translations:

http://guides.rubyonrails.org/i18n.html#using-safe-html-translations

lewispb:
@Cruikshanks Just thinking about this, should the copyright year be dynamic ?

Cruikshanks:
No I don't believe so. You fix it at the point you apply the right. It's similar to what we do in LICENSE files; we'd only change the copyright date if we ever changed the license.

PR #250, opened 2016-12-20
Update version due to OS place T&C's change

Cruikshanks:
We the adition of the OS Places notice and T&C's to the engine (see a1e3e09), we need to update the version and tag the engine, so we can then update the [front office](https://github.com/EnvironmentAgency/flood-risk-front-office) and [back office](https://github.com/EnvironmentAgency/flood-risk-back-office) to bring in the changes.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    flood_risk_engine (1.0.1)
+    flood_risk_engine (1.0.2)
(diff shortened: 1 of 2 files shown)


Cruikshanks:
Sorry chaps, have broken something. Will ask for the review once I've got it fixed.

Cruikshanks:
Ok, solved it. Who new once you update the version you also need to run `bundle` and ensure the `Gmfile.lock` is update appropriately.

PR #256, opened 2018-03-16
Upgrade mailcatcher to resolve CVE-2018-7212

Cruikshanks:
The previous version of Mailcatcher had a dependency on Sinatra 1.4.8 that our CI tools were flagging because of CVE-2018-7212.

Updating the version of mailcatcher we use should resolve this.


Cruikshanks:
Doesn't actually fix the problem and looks like this will need more work so closing for now.

PR #257, opened 2019-05-31
Update I18n

irisfaraway:
When we were setting up the vagrant box for FRAE, we noticed the `bundle install` was breaking. Updating the dependency solves the problem.


irisfaraway:
https://github.com/DEFRA/flood-risk-engine/pull/259 covers this

PR #258, opened 2019-05-31
Update privacy policy

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-293

This also makes some changes to standardise how we present privacy policies with WCR and WEX – we usually just hardcode the text instead of using I18n for static pages, and we also have some additional styling for the date modified metadata at the bottom.

This date will need to be updated once we have a release date scheduled.


irisfaraway:
Hakiri still failing but I'm dealing with that on a separate branch.

PR #260, opened 2019-06-06
Update dependencies

irisfaraway:
We had a lot of out-of-date gems, which means security vulnerabilities. This PR will cover those updates.

I decided to lock Rubocop to the existing version as updating it would have raised a lot of linter issues. Long-term we should switch to using https://github.com/DEFRA/defra-ruby-style but that'll need a separate PR (or several).


irisfaraway:
Good news: Hakiri is green! Bad news: the build is broken again!

irisfaraway:
source

PR #291, opened 2019-08-20
Migrate and fix area lookup to defra-ruby-area gem

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-484

The EA administrative boundaries Web Feature Services were transferred to a different supplier in December 2018, which also seemed to break the existing API.

We've finally got a confirmed working API in June 2019. But rather than update the existing [ea-area_lookup](https://github.com/DEFRA/ea-area_lookup) gem we decided instead to start afresh and create [defra-ruby-area](https://github.com/DEFRA/defra-ruby-area).

This update migrates the engine to use the new gem, which essentially means updating how it makes the call and then handles the response.


Cruikshanks:
I agree with CodeClimate regards the 2 issues. However they are existing issues simply re-flagged because I've touched the code! So I'd argue to leave as is for now.

And regards the drop in test coverage I'm not sure where it it. I'm still seeing the `UpdateWaterManagementAreaJob` and model `WaterManagementArea` having 100% test coverage, so I'm just ignoring that complaint for now.

app/jobs/flood_risk_engine/update_water_management_area_job.rb
```diff
@@ -12,29 +12,29 @@ def perform(location)
...
-                                               northing: location.northing)
-      EA::AreaLookup.find_water_management_area_by_coordinates(coords)
+      process_unsucessful_response(response)
     end
 
-    def find_or_create_area_from_api_result(api_result)
-      WaterManagementArea.find_or_create_by(code: api_result[:code]) do |area|
-        area.update_attributes(api_result)
+    def process_successful_response(result)
+      WaterManagementArea.find_or_create_by(code: result.code) do |area|
```
cintamani: Have you heard about https://medium.com/aviabird/ruby-tap-that-method-90c8a801fd6a before? feels like it can be useful here :)

spec/jobs/flood_risk_engine/update_water_management_area_spec.rb
```diff
@@ -1,5 +1,17 @@
...
+  module Area
+    Response = Struct.new(:area, :successful, :error) do
+      def successful?
+        successful
+      end
+    end
+
+    Area = Struct.new(:area_id, :area_name, :code, :long_name, :short_name)
+  end
+end
```
cintamani: Can we move this in a support file?

irisfaraway:
@Cruikshanks I tried running the tests locally on this branch and master. There are 4 fewer "relevant lines", all of which had test coverage, so the proportion of covered lines has dropped slightly. But we haven't actually lost any coverage at all 👍

app/jobs/flood_risk_engine/update_water_management_area_job.rb
Cruikshanks: That's a bit of ruby I never knew about 😁 🤯

However I'm not sure I've got how you think it can help here. I've tested locally with this

```ruby
WaterManagementArea.find_or_create_by(code: result.code).tap do |area|
  area.update_attributes(
    area_id: result.area_id,
    area_name: result.area_name,
    short_name: result.short_name,
    long_name: result.long_name
  )
end
```

But that has just added `.tap` into some code that previously didn't need it. Would be grateful for some advice on what I should be doing.

spec/jobs/flood_risk_engine/update_water_management_area_spec.rb
Cruikshanks: I can do, but think I'd rather do it as a separate housekeeping exercise. I got the idea of using a struct from stuff we've done with the validators, for example https://github.com/DEFRA/defra-ruby-validators/blob/master/spec/defra_ruby/validators/business_type_validator_spec.rb

In all the validator specs, and those that are still left in [waste-exemptions-engine](https://github.com/DEFRA/waste-exemptions-engine) the struct is declared at the top of the spec.

So if we'd rather these live in `spec/support`, I'd push for cleaning them all up at once, and recording the decision. In which case I can raise an issue in [ruby-services-team](https://github.com/DEFRA/ruby-services-team) to log and track the task??

app/jobs/flood_risk_engine/update_water_management_area_job.rb
cintamani: Is quite recent stuff :) And yes, you are right, in here is actually not useful as `find_or_create_by` behaves like tap already.

spec/jobs/flood_risk_engine/update_water_management_area_spec.rb
cintamani: Sounds good :)

Cruikshanks: Created https://github.com/DEFRA/ruby-services-team/issues/65

PR #297, opened 2019-09-03
Update water mgmt area job to report errors

Cruikshanks:
Prior to this change if an error occurred when the engine attempted to lookup the water mangement area for an exemption nothing was logged.

This would mean we on the service team would no longer be aware of any issues with the WFS service until the users reported it.

This change adds reporting of errors to Errbit, and some tweaks to the job class to support it.

### DIFF
--- a/app/jobs/flood_risk_engine/update_water_management_area_job.rb
+++ b/app/jobs/flood_risk_engine/update_water_management_area_job.rb
@@ -3,16 +3,18 @@ module FloodRiskEngine
     def perform(location)
       raise MissingLocationArgumentError unless location
 
+      @location = location
+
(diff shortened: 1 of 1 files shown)


Cruikshanks:
Regards the codeclimate issue, I would like to mark it as `wontfix`. In CodeClimate's documentation all the examples assume access to an initializer, where the instance variable is then set.

Its solution is to simply replace the instance variable with an attr_reader. As we are in a job I don't believe the examples or the suggested solution is appropriate or needed.

`perform()` is the one public method on the instance and it expects `location` to be provided, so I believe it is fine to then assume `@location` has been set in all subsequent methods.

PR #323, opened 2020-01-24
Move base service to correct directory

cintamani:
🙈


cintamani:
I suppose the same I was doing when I wrote and opened it 😂

PR #332, opened 2020-02-19
Add defra-ruby-email to project

Cruikshanks:
Currently the [Waste Exemptions service](https://github.com/DEFRA/waste-exemptions-engine/pull/112) and the [Waste Carriers Frontend](https://github.com/DEFRA/waste-carriers-frontend/pull/227) have the ability to intercept and play back the details of the last email sent.

This feature is only enabled in our non-production environments and is used as part of our acceptance tests. However our QA @andrewhick has rightly pointed out that the functionality is inconsistent across our services.

Waste Exemptions has it throughout, Waste Carriers only has it in the old app, and Flood Risk Activity Exemptions doesn't have it at all. This makes writing and maintaining acceptance tests across the 3 services difficult and inconsistent so we have been asked to resolve the issue.

Rather than duplicate the existing code even more, we have created [defra-ruby-email](https://github.com/DEFRA/defra-ruby-email) a reusable engine that can be mounted into an application to provide the same functionality.

Having created the gem, this change covers adding it to FRAE so it too can benefit from the feature.

### DIFF
--- a/lib/flood_risk_engine/engine.rb
+++ b/lib/flood_risk_engine/engine.rb
@@ -4,6 +4,7 @@ require "activerecord/session_store"
 require "high_voltage"
 require "ea/address_lookup"
 require "defra_ruby/area"
+require "defra_ruby_email"
 require "active_support/dependencies"
 require_dependency "virtus"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,6 +5,7 @@ PATH
       activerecord-session_store (~> 1.0)
       defra_ruby_alert (~> 0.1.0)
       defra_ruby_area
+      defra_ruby_email
       dibber (~> 0.5)
       dotenv-rails (~> 2.1)
       ea-address_lookup (~> 0.3.0)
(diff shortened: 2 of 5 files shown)


Cruikshanks:
Please CodeClimate, it's such a teeny-tiny change. Can we agree to overlook it this one time? 😄 😉

lib/flood_risk_engine/engine.rb
```diff
@@ -4,6 +4,7 @@
 require "high_voltage"
 require "ea/address_lookup"
 require "defra_ruby/area"
+require "defra_ruby_email"
```
irisfaraway: Is there a reason we have `defra_ruby_email` and not `defra_ruby/email` like we do for our other gems?

Cruikshanks: There seems to be a difference in how namespacing is handled in an engine compared to a plain old gem.

I couldn't get to the bottom of it, so fell back to what rails generates. We had the same issue in [defra-ruby-mocks](https://github.com/DEFRA/defra-ruby-mocks) hence it too has a `defra_ruby_mocks` namespace.

irisfaraway: Ahh, interesting, I didn't know that! Good to know 👍

PR #376, opened 2021-05-18
Upgrade engine to Rails 6.0 and Ruby 2.7

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1396


spec/factories/organisations_factory.rb
```diff
@@ -12,9 +12,12 @@
     # One trait per type named in format : as_xxx e.g   as_local_authority, as_limited_liability_partnership etc
     FloodRiskEngine::Organisation.org_types.each_key do |ot|
       trait :"as_#{ot}" do
-        org_type ot.to_s
-        name ""
-        registration_number ""
+        org_type            { ot.to_s }
+        name                { "" }
+        # this errors with `unknown attribute`
```
irisfaraway, line 17: If this is no longer required, I'd rather we remove it entirely than just leave it commented out.

db/migrate/20160415200213_add_missing_indexes.rb
```diff
@@ -1,12 +1,11 @@
-class AddMissingIndexes < ActiveRecord::Migration
+class AddMissingIndexes < ActiveRecord::Migration[5.2]
   def change
     add_index :flood_risk_engine_enrollments, :applicant_contact_id
     add_index :flood_risk_engine_enrollments, :organisation_id
     add_index :flood_risk_engine_enrollments, :site_address_id
     add_index :flood_risk_engine_contacts,
               :partnership_organisation_id,
               name: 'fre_contacts_partnership_organisation_id'
-    add_index :flood_risk_engine_organisations, :contact_id
```
irisfaraway, line 9: Why have we removed this?

db/migrate/20160415200213_add_missing_indexes.rb
```diff
@@ -1,12 +1,11 @@
-class AddMissingIndexes < ActiveRecord::Migration
+class AddMissingIndexes < ActiveRecord::Migration[5.2]
```
irisfaraway, line 1: If these migrations were generated in Rails 4, should we have `[5.2]` here or should it be `[4.2]`? Had a look and that's what we did in WEX.

irisfaraway (commented):
This is looking good 👏 Also great to see the test suite running without one million deprecation warnings! Will make it so much less painful.

I have a few questions but nothing major.

db/migrate/20160415200213_add_missing_indexes.rb
tobyprivett: There was a duplicate index error when I ran the migrations. 
The `contact_id` is added as a reference column which should add the index:
https://github.com/DEFRA/flood-risk-engine/blob/c72541ddfe582426ee7a1dfcb2d313b2d324ab8d/db/migrate/20160406132213_create_flood_risk_engine_organisations.rb#L6

Just ran the migrations again locally, and checked in postgres for the index: ` "index_flood_risk_engine_organisations_on_contact_id" btree (contact_id)`,

spec/factories/organisations_factory.rb
tobyprivett: Removed! :-)

irisfaraway (approved):
:shipit:

db/migrate/20160415200213_add_missing_indexes.rb
tobyprivett: Done!

PR #381, opened 2021-06-10
Ensure user selects an exemption

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1469

- Something to do with the params have changed since the Rails upgrade,
and the validator (add_exemption_form.rb) thinks that
{ exemption_ids: "" } is valid

- The validator specifies min length of 1, but this is voodoo (Reform gem)
validation and it seems to treat the hash as having a length of 1

- This 'fix' removes any param keys without values and seems to 'work'

### DIFF
--- a/app/forms/flood_risk_engine/steps/add_exemptions_form.rb
+++ b/app/forms/flood_risk_engine/steps/add_exemptions_form.rb
@@ -19,6 +19,19 @@ module FloodRiskEngine
         @all_exemptions = Exemption.all
       end
 
+      def validate(params)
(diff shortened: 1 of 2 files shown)


tobyprivett:
@irisfaraway This is a bit of form validation weirdness. It took a while hunt down and there may be a better way of doing it. What are your thoughts?

tobyprivett:
This PR might not be necessary if we can get the refactor work done

tobyprivett:
Closing this PR because we will be fixing the bug as part of the refactor work

PR #384, opened 2021-06-14
Return correct error for valid postcode without results

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1341

Previously if we searched for a postcode with no results, it would display an error message saying that the address finder was not working. This PR fixes the issue so we instead say that no results were found.

I had a look at the spec for this and it was wrong, as it was trying to raise a 400 error which is no longer the response we expect, and then testing for the wrong error message anyway.

### DIFF
--- a/app/forms/flood_risk_engine/steps/base_address_search_form.rb
+++ b/app/forms/flood_risk_engine/steps/base_address_search_form.rb
@@ -46,7 +46,7 @@ module FloodRiskEngine
 
         return true if response.successful?
 
-        if response.error == DefraRuby::Address::NoMatchError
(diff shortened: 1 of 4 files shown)


app/forms/flood_risk_engine/steps/base_address_search_form.rb
```diff
@@ -46,7 +46,7 @@ def look_up_addresses
 
         return true if response.successful?
 
-        if response.error == DefraRuby::Address::NoMatchError
+        if response.error.is_a?(DefraRuby::Address::NoMatchError)
```
tobyprivett, line 49: Much better than `==`!

tobyprivett (approved):

PR #385, opened 2021-06-15
Clean up duplicate address after address is changed

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1341

During testing we discovered an issue where if an address was added via postcode lookup, and then changed later to one entered manually, the previous address would still be assigned and often displayed instead of the amended, manually-entered version.

To fix this, I've added a callback to clean up any duplicate addresses after a new one is added. The intent is to refactor these forms soon anyway so this is a temporary fix.

### DIFF
--- a/app/models/flood_risk_engine/address.rb
+++ b/app/models/flood_risk_engine/address.rb
@@ -39,9 +39,10 @@ module FloodRiskEngine
     private
 
     def clean_up_duplicate_addresses
-      return unless addressable.is_a?(FloodRiskEngine::Organisation)
+      return unless addressable.is_a?(FloodRiskEngine::Organisation) && address_type == "primary"
 
-      primary_addresses = FloodRiskEngine::Address.where(addressable: addressable)
+      primary_addresses = FloodRiskEngine::Address.where(addressable: addressable,
+                                                         address_type: "primary")
 
       primary_addresses.each do |address|
         next if address == self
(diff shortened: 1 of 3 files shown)


app/models/flood_risk_engine/address.rb
```diff
@@ -33,5 +35,18 @@ def parts
     def address_methods
       %i[premises street_address locality city postcode]
     end
+
+    private
+
+    def clean_up_duplicate_addresses
+      return unless addressable.is_a?(FloodRiskEngine::Organisation)
+
+      primary_addresses = FloodRiskEngine::Address.where(addressable: addressable)
```
tobyprivett, line 44: Will this delete all the addresses for this organisation? 'Primary' suggests there might be 'secondary', etc. - or is that not relevant?

tobyprivett (approved):

app/models/flood_risk_engine/address.rb
irisfaraway: There is code for 9(!) types of address:

```
      primary: 0,
      secondary: 1,
      site: 2,
      contact: 3,
      billing: 4,
      principal_office: 5,
      principal_business: 6,
      registered_office: 7,
      partner: 8
```

But from looking at the data, the only one we actually ever seem to use is the primary one – I think a lot of this was in there in case the db was ever hooked up with the CRM. We definitely don't have a need for a billing address in a free service, for example. The only other one that there's evidence of use is the contact address, although there doesn't appear to be a way to add that any more, and the addressable class for that is the Contact. So I think we're OK but I'll add the `address_type` check in as well just to be safe.

tobyprivett: I think there was a hefty cut-and-paste when this app was built!

tobyprivett (approved):

PR #391, opened 2021-06-21
Add Simplecov JSON formatter for test coverage in Sonarcloud

irisfaraway:
As part of the Rails 6 upgrade we updated Simplecov from 0.13 to 0.21. During these version jumps, simplecov got rid of the JSON format output as standard and it instead has to be added through a separate gem. Without this gem, we were unable to export our test coverage to Sonarcloud and it continually reported that we had 0% test coverage.


irisfaraway:
Binning this for now and just pinning to an earlier version instead.

PR #393, opened 2021-06-22
Set up controllers for updated new reg workflow

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1472

This PR sets up new controllers and routes for the updated new registration workflow. These will follow the format we already use in WCR and WEX.

The end result should be an empty 'happy path' journey that a user can click through and see the pages in the expected journey. The actual content of the pages won't be merged in yet.

### DIFF
--- a/db/migrate/20210624113422_add_type_to_transient_registrations.rb
+++ b/db/migrate/20210624113422_add_type_to_transient_registrations.rb
@@ -1,0 +1,14 @@
+class AddTypeToTransientRegistrations < ActiveRecord::Migration[6.0]
+  def up
+    add_column :transient_registrations, :type, :string, null: false
+    # Set any existing TransientRegistrations to NewRegistrations so everything has a type
+    execute <<-SQL
+       UPDATE transient_registrations
+       SET type = 'FloodRiskEngine::NewRegistration';
+    SQL
+  end
+
+  def down
+    remove_column :transient_registrations, :type
+  end
+end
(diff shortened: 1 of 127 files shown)


db/migrate/20210624113422_add_type_to_transient_registrations.rb
```diff
@@ -0,0 +1,14 @@
+class AddTypeToTransientRegistrations < ActiveRecord::Migration[6.0]
+  def up
+    add_column :transient_registrations, :type, :string, null: false
+    # Set any existing TransientRegistrations to NewRegistrations so everything has a type
+    execute <<-SQL
+       UPDATE transient_registrations
+       SET type = 'FloodRiskEngine::NewRegistration';
+    SQL
+  end
```
tobyprivett, line 9: This didn't work for me locally, as the type column doesn't allow nulls. Perhaps add the column with a default value, then add the `null:false` afterwards?

tobyprivett: This could be a development only issue as the transient_registrations table is new.

tobyprivett (commented):
Looks good!
We'll have quite a lot of work updating the html templates and i18n.

Just a comment about set up woes with the `type` column. I think it needs re-jigging.

db/migrate/20210624113422_add_type_to_transient_registrations.rb
irisfaraway: I've had a go at setting up a default first and then adding null later – how's this work for you? https://github.com/DEFRA/flood-risk-engine/pull/393/commits/e527647807d636d4752d0c1569ddc9c1aaaf70d2

tobyprivett: Looks good! Is there an index on the `type`?

irisfaraway: There isn't in WEX so I didn't add one here, but might be a good idea to add it

PR #394, opened 2021-07-05
Remove erroneous `pages` route

tobyprivett:
- A `pages` route is declared earlier in the file and points to the
`enrollments/pages` controller

- Here we remove the additional `pages` route as it points to the
wrong controller


irisfaraway (approved):
Oops, my bad, I added this by mistake and then thought I'd removed it but clearly not 🤦‍♀️

PR #396, opened 2021-07-12
Add validation and params to new workflow forms

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1473

This PR sets up the models and migrations for the various transient objects. As well as the transient registration, we also have an address, person and registration_exemption. Unlike WEX and WCR, in this case addresses can belong to partners rather than to the organisation, so we use polymorphic association.

It also includes form validations for these forms:

- exemption
- business type
- company name
- company number
- company postcode
- contact name
- contact phone
- contact email
- additional contact email
- declaration

Not included are the rest of the company address pages, the site location and the partner pages.

### DIFF
--- a/db/migrate/20210629070929_create_transient_addresses.rb
+++ b/db/migrate/20210629070929_create_transient_addresses.rb
@@ -20,6 +20,9 @@ class CreateTransientAddresses < ActiveRecord::Migration[6.0]
       t.string :addressable_type
       t.string :uprn
       t.string :token
+
+      t.index :addressable_id, unique: true
+      t.index :addressable_type, unique: true
     end
   end
 end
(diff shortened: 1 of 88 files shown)


db/migrate/20210629070929_create_transient_addresses.rb
```diff
@@ -0,0 +1,25 @@
...
+      t.string :postcode, limit: 8
+      t.integer :county_province_id
+      t.string :country_iso, limit: 3
+      t.integer :address_type, default: 0, null: false
+      t.string :organisation, limit: 255, default: ""
+      t.date :state_date
+      t.string :blpu_state_code
+      t.string :postal_address_code
+      t.string :logical_status_code
+      t.integer :addressable_id
```
tobyprivett, line 19: The foreign keys will need an index

tobyprivett (approved):
Just a comment about indexing the foreign keys (I've been caught out by this before so I always have an eye on potential issues)

Looks like good, so will approve and let you crack on!

db/migrate/20210629070929_create_transient_addresses.rb
irisfaraway: Good spot, will sort it!

PR #401, opened 2021-08-02
Add content to exemptions in the new workflow

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1460

This PR adds content for the forms to select and confirm an exemption, and saves the submitted data.

It also updates the styling for the back link.

---

Depends on https://github.com/DEFRA/flood-risk-engine/pull/400

PR #402, opened 2021-08-02
Add content to site location page in new workflow

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1461

This PR adds the form content and validation for the site location page.

It also includes a migration to add an additional field to the transient registration for the dredging length, which is only used for one exemption.

---

Depends on https://github.com/DEFRA/flood-risk-engine/pull/401

### DIFF
--- a/db/migrate/20210802151907_add_dredging_length_to_transient_registration.rb
+++ b/db/migrate/20210802151907_add_dredging_length_to_transient_registration.rb
@@ -1,0 +1,5 @@
+class AddDredgingLengthToTransientRegistration < ActiveRecord::Migration[6.0]
+  def change
+    add_column :transient_registrations, :dredging_length, :integer
+  end
+end
--- a/lib/flood_risk_engine/configuration.rb
+++ b/lib/flood_risk_engine/configuration.rb
@@ -32,7 +32,8 @@ module FloodRiskEngine
 
     config_accessor(:redirection_url_postcode_lookup)
     config_accessor(:layout) { "application" }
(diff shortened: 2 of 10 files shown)


lib/flood_risk_engine/configuration.rb
```diff
@@ -32,7 +32,8 @@ class Configuration
 
     config_accessor(:redirection_url_postcode_lookup)
     config_accessor(:layout) { "application" }
-    config_accessor(:minumum_dredging_length_in_metres) { 1 }
+    config_accessor(:minumum_dredging_length_in_metres) { 1 } # TYPO - left in for legacy purposes
```
tobyprivett, line 35: Good to fix that!

db/migrate/20210802151907_add_dredging_length_to_transient_registration.rb
```diff
@@ -0,0 +1,5 @@
+class AddDredgingLengthToTransientRegistration < ActiveRecord::Migration[6.0]
+  def change
+    add_column :transient_registrations, :dredging_length, :string
```
tobyprivett, line 3: This is validated as an integer. Could it be an integer column also?

tobyprivett (approved):
Looking good! Just a comment about the data type for dredging_length.

db/migrate/20210802151907_add_dredging_length_to_transient_registration.rb
irisfaraway: Yeah, for some reason it was a string in the existing records but I really don't see why! I'll change it 👍

PR #405, opened 2021-08-03
Add content for company info pages in new workflow

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1462

This includes:

- business type
- business name
- registered Companies House number (if valid)
- company address
- partner details

### DIFF
--- a/app/forms/concerns/flood_risk_engine/can_update_partner.rb
+++ b/app/forms/concerns/flood_risk_engine/can_update_partner.rb
@@ -12,10 +12,7 @@ module FloodRiskEngine
       private
 
       def setup_partner
-        puts "AWOOOOOOOGA"
-        puts transient_registration.transient_people.to_json
         self.partner = transient_registration.transient_people.last
-        puts partner.to_json
       end
     end
   end
--- a/app/forms/concerns/flood_risk_engine/can_update_partner.rb
+++ b/app/forms/concerns/flood_risk_engine/can_update_partner.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module FloodRiskEngine
+  module CanUpdatePartner
+    extend ActiveSupport::Concern
+
+    included do
+      attr_accessor :partner
+
+      after_initialize :setup_partner
+
+      private
+
+      def setup_partner
+        puts "AWOOOOOOOGA"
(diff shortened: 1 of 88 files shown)


app/forms/concerns/flood_risk_engine/can_update_partner.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    included do
+      attr_accessor :partner
+
+      after_initialize :setup_partner
+
+      private
+
+      def setup_partner
+        puts "AWOOOOOOOGA"
```
tobyprivett, line 15: Perhaps remove the debug code?

app/views/flood_risk_engine/shared/_manual_address.html.erb
```diff
@@ -0,0 +1,46 @@
...
+
+  <%= f.govuk_text_field :premises,
+      label: { text: t(".premises_label") },
+      hint: { text: t(".premises_hint") },
+      width: "one-half",
+      value: form.premises
+  %>
+</div>
+
+<div class="govuk-form-group <%= "govuk-form-group--error" if form.errors[:street_address].any? %>">
```
tobyprivett, line 14: Do we have access to `form` here? It might just be `f`. The errors should show up with the `govuk_text_field` helper, so the form-group wrapper may not be needed. Might be worth having another look

tobyprivett (approved):
Looking good! One comment about debug `puts`. And a suggestion about using both references form objects in the same view. Hope the latter makes sense, happy to discuss.

app/forms/concerns/flood_risk_engine/can_update_partner.rb
irisfaraway: Hahaha oops! Will get rid!

app/views/flood_risk_engine/shared/_manual_address.html.erb
irisfaraway: In this case `form` is representing the form instance variable, eg `@partner_manual_address_form`, so that's why it's different to `f`. There aren't really any other good examples of it in FRAE but we have similar things in WCR, for example: https://github.com/DEFRA/waste-carriers-engine/blob/main/app/views/waste_carriers_engine/shared/_dob.html.erb

PR #418, opened 2021-09-14
Delete code from old registration journey

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1592

Now that we have the new registration journey working with AASM, we no longer need the old one that uses Reform and finite_machine. So this PR removes a whole bunch of old code around that.


tobyprivett (approved):
I checked out the branch and it looks like you've covered everything. A big improvement!!!

PR #423, opened 2021-09-17
Fix issues on company details pages

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1462

This PR fixes a range of technical and design issues found during QA and design review.

### DIFF
--- a/app/views/flood_risk_engine/company_address_manual_forms/new.html.erb
+++ b/app/views/flood_risk_engine/company_address_manual_forms/new.html.erb
@@ -4,6 +4,7 @@
   <div class="govuk-grid-column-two-thirds">
     <%= form_for @company_address_manual_form do |f| %>
       <%= render partial: "flood_risk_engine/shared/error_summary", locals: { f: f } %>
+
       <% if @company_address_manual_form.address_finder_error %>
         <div class="govuk-error-summary">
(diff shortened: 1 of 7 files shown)


app/views/flood_risk_engine/company_address_manual_forms/new.html.erb
```diff
@@ -20,15 +21,17 @@
 
       <p class="govuk-body">
         <%= t(".preset_postcode_label") %>
-      <br/>
-      <span class="postcode govuk-!-font-weight-bold">
-        <%= @company_address_manual_form.postcode %>
-      </span>
-      <%= link_to(t(".postcode_change_link"), back_company_address_manual_forms_path(@company_address_manual_form.token)) %>
+        <br/>
+        <span class="postcode govuk-!-font-weight-bold">
```
tobyprivett, line 25: I don't think we have a css class for '.postcode' anymore, so that can probably go?

tobyprivett (approved):
Just a note about the postcode css

app/views/flood_risk_engine/company_address_manual_forms/new.html.erb
irisfaraway: We do in the host apps – it adds a bit of extra padding between the postcode and the change link

PR #433, opened 2021-09-20
Fix version page in production mode

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1481

The old GitCommitSha method of getting the current commit worked fine in local development environments but threw an error when running in production mode. This PR cribs the fix we use in WCR and WEX to get the version from Capistrano.

### DIFF
--- a/app/presenters/flood_risk_engine/application_version_page_presenter.rb
+++ b/app/presenters/flood_risk_engine/application_version_page_presenter.rb
@@ -6,10 +6,6 @@ module FloodRiskEngine
       FloodRiskEngine.config.application_name || Rails.application.class.module_parent_name
     end
 
-    def application_version
-      defined?(::Application::VERSION) ? Application::VERSION : "Undefined"
-    end
-
     def git_commit
       @_git_commit ||= begin
         sha =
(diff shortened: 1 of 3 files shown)


app/presenters/flood_risk_engine/application_version_page_presenter.rb
```diff
@@ -6,12 +6,24 @@ def application_name
...
+        sha =
+          if Rails.env.development?
+            `git rev-parse HEAD`
+          else
+            heroku_file = Rails.root.join ".source_version"
+            capistrano_file = Rails.root.join "REVISION"
+
+            if File.exist? capistrano_file
+              File.open(capistrano_file, &:gets)
+            elsif File.exist? heroku_file
```
tobyprivett, line 20: Are we using heroku?

tobyprivett (approved):

app/presenters/flood_risk_engine/application_version_page_presenter.rb
irisfaraway: I think it may occasionally get used for testing, but it's not in active use right now.

PR #500, opened 2022-06-15
[1910] Update ruby to 3.1.2

d-a-v-e:
https://eaflood.atlassian.net/browse/RUBY-1910


.rubocop.yml
```diff
@@ -121,8 +124,8 @@ Style/MethodCalledOnDoEndBlock:
 
 # Rubocop 0.42 introduced this cop - disabling for now until we have time to resolve
 # the issues it raises.
-Style/MethodMissing:
-  Enabled: false
+# Style/MethodMissing:
```
PaulDoyle-DEFRA, line 127: Remove if not required

.rubocop_todo.yml
```diff
@@ -0,0 +1,77 @@
+# This configuration was generated by
```
PaulDoyle-DEFRA, line 1: How does this config relate to the standard DEFRA config in the defra_ruby_style gem?

app/controllers/flood_risk_engine/start_forms_controller.rb
```diff
@@ -17,7 +17,7 @@ def transient_registration_attributes
     end
 
     def find_or_initialize_transient_registration(token)
-      @transient_registration = NewRegistration.where(token: token).first ||
+      @transient_registration = NewRegistration.where(token:).first ||
```
PaulDoyle-DEFRA, line 20: Not sure I like this new Rubocop recommendation, might be worth discussing whether to disable it globally.

flood_risk_engine.gemspec
```diff
@@ -1,69 +1,73 @@
...
+  s.authors               = ["Digital Services Team, EnvironmentAgency"]
+  s.email                 = ["dst@environment-agency.gov.uk"]
+  s.homepage              = "https://github.com/defra/flood-risk-engine"
+  s.summary               = "FloodRiskEngine package containing core elements and functionality"
+  s.description           = "FloodRiskEngine package containing core elements and functionality"
+  s.license               = "The Open Government Licence (OGL) Version 3"
+  s.required_ruby_version = '>= 3.1'
 
   s.files = Dir["{app,config,db,lib}/**/*", "LICENSE", "Rakefile", "README.rdoc"]
-  s.test_files = Dir["spec/**/*"]
```
PaulDoyle-DEFRA, line 18: Why has this been commented out?

spec/rails_helper.rb
```diff
@@ -5,7 +5,6 @@
 
 # Prevent database truncation if the environment is production
 abort("The Rails environment is running in production mode!") if Rails.env.production?
-require "spec_helper"
```
PaulDoyle-DEFRA, line 8: Why has this been removed?

flood_risk_engine.gemspec
d-a-v-e: this seems to be a deprecated config: https://docs.rubocop.org/rubocop/cops_gemspec.html#gemspecdeprecatedattributeassignment

spec/rails_helper.rb
d-a-v-e: it's already being required up at the top - think it was added by accident in the past

.rubocop_todo.yml
d-a-v-e: the gem doesn't seem to actually use `defra_ruby_style`. I generated this using rubocop after enabling the latest cops  via `NewCops: enable` in `.rubocop.yml`. This is because the fixes it was suggesting would have (potentially) broken the engine.

app/controllers/flood_risk_engine/start_forms_controller.rb
PaulDoyle-DEFRA: I just think the old format was easier to scan. Since you've done the work here let's leave it for now, but let's review before the other projects are migrated to Ruby v3+.

PaulDoyle-DEFRA (approved):

PR #505, opened 2022-06-22
[1910] Add defra ruby style gem

d-a-v-e:
https://eaflood.atlassian.net/browse/RUBY-1910


.rubocop.yml
```diff
@@ -126,3 +130,6 @@ Style/MethodCalledOnDoEndBlock:
 # value in using "" for all strings irrespective of whether string interpolation is used
 Style/StringLiterals:
   EnforcedStyle: double_quotes
+
+Style/HashSyntax:
+  Enabled: false
```
PaulDoyle-DEFRA, line 135: I think this should be enabled in the master style within the gem and not overruled here.

PaulDoyle-DEFRA (approved):

.rubocop.yml
d-a-v-e: ah yeah, it is already! I can remove this. thanks

PR #721, opened 2024-01-10
Fix/ruby 2805 postcode validation

PaulDoyle-DEFRA:
Change to use the new postcode validator from defra_ruby_validators
https://eaflood.atlassian.net/browse/RUBY-2805


flood_risk_engine.gemspec
```diff
@@ -36,6 +36,7 @@ Gem::Specification.new do |s|
   # which details of the last email sent by the app can be accessed
   s.add_dependency "defra_ruby_email", "~> 1"
   s.add_dependency "defra_ruby_validators"
+  s.add_dependency "faraday-retry"
```
brujeo, line 39: Shouldn't this be a Gemfile dependency?

spec/rails_helper.rb
```diff
@@ -53,12 +53,12 @@
...
 
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
-  config.fixture_path = Rails.root.join("/spec/fixtures")
+  # config.fixture_path = Rails.root.join("/spec/fixtures")
 
   # If you're not using ActiveRecord, or you'd prefer not to run each of your
   # examples within a transaction, remove the following line or assign false
   # instead of true.
-  config.use_transactional_fixtures = true
+  config.use_transactional_fixtures = false
```
brujeo, line 61: Just wondering what the motivation was for setting this to false? As far as I'm aware transactional fixtures are way faster than the use of DB cleaner after each test

flood_risk_engine.gemspec
PaulDoyle-DEFRA: Perhaps. Do you mean just for :development?

brujeo: in my understanding, faraday is only used by changelog gem, which isn't used in the runtime

spec/rails_helper.rb
PaulDoyle-DEFRA: As discussed separately - I found the cause of the test failures and have pushed a commit with a fix, and reinstated config.use_transactional_fixtures = true

brujeo (approved):

PR #772, opened 2024-04-16
Fix/ruby 3027 area lookup

PaulDoyle-DEFRA:
This change is part of a move away from calling a remote and undocumented API to look up Water Management Areas and instead looking up an internally hosted spatial dataset using rgeo.


.github/workflows/brakeman-analysis.yml
```diff
@@ -20,7 +20,7 @@ on:
 jobs:
   brakeman-scan:
     name: Brakeman Scan
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
brujeo, line 23: Any particular reason to stick with a specific version of ubuntu?

brujeo (dismissed):

brujeo (dismissed):

brujeo (approved):

PR #781, opened 2024-04-30
Cleanup

PaulDoyle-DEFRA:
Minor gem updates and general tidy-up.


Gemfile.lock
```diff
@@ -513,6 +515,7 @@ GEM
     zeitwerk (2.6.13)
 
 PLATFORMS
+  x86_64-darwin-21
   x86_64-linux
```
brujeo, line 519: This doesn't look right ... In normal circumstances, platform should be set to :ruby

PaulDoyle-DEFRA: Agreed - not sure how that crept in. Fixed now.

brujeo (approved):

PR #782, opened 2024-04-30
Fix/easting northing validation

PaulDoyle-DEFRA:
This fixes an issue where easting/northing strings with a leading zero were being interpreted as octal by Ruby.


spec/services/flood_risk_engine/water_management_area_lookup_service_spec.rb
brujeo, line 1: I would expect a test use case with easting/northing leading zero's

brujeo (approved):

spec/services/flood_risk_engine/water_management_area_lookup_service_spec.rb
PaulDoyle-DEFRA: See line 54

PR #1084, opened 2026-05-18
Update bundle dependencies

jjromeo:
Run bundle update


Gemfile.lock
```diff
@@ -552,7 +584,19 @@ GEM
...
+  arm-linux-gnu
+  arm-linux-musl
+  arm64-darwin
   ruby
+  x86-linux-gnu
+  x86-linux-musl
+  x86_64-darwin
+  x86_64-linux
+  x86_64-linux-gnu
+  x86_64-linux-musl
```
brujeo: Why do we need all these? Should be just ruby here

brujeo (approved):

